PR #7, opened 2017-09-08
Fix imports for python3

SleuthKid:
This PR fixes import issues for configparser when building the tool with python3.

### DIFF
--- a/bro-pkg
+++ b/bro-pkg
@@ -19,7 +19,10 @@ import filecmp
 import shutil
 import subprocess
 
-from backports import configparser
+try:
+    from backports import configparser
+except ImportError as err:
+    import configparser
 
 import bropkg
 
--- a/bropkg/manager.py
+++ b/bropkg/manager.py
@@ -11,7 +11,11 @@ import filecmp
 import tarfile
 import subprocess
 
-from backports import configparser
+try:
+    from backports import configparser
+except ImportError as err:
+    import configparser
+
 import git
 import semantic_version as semver
 
--- a/bropkg/source.py
+++ b/bropkg/source.py
@@ -10,7 +10,10 @@ import os
 import shutil
 import git
 
-from backports import configparser
+try:
+    from backports import configparser
+except ImportError as err:
+    import configparser
 
 from . import LOG
 from .package import (


jsiwek:
Can you elaborate on the actual problem you're running in to?

The code as it stands was meant to work on both python2 and python3, provided that python is able to find the backported version of configparser: https://pypi.python.org/pypi/configparser

Installing via the suggested method of `pip` should automatically install that along with other dependencies (the 'install_requires' field in setup.py).

SleuthKid:
I am currently packaging this repo for the Debian project (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=874704).

Deb packaging does not allow pip dependencies and requires to use the Debian distribution python packages as build dependencies. I had to use the patch in this PR to get it working.

hillu:
Python 3.5 and higher already contain `configparser`, so there is no need to use or even install the "backports" version. See #8 which includes @SleuthKid's PR.

jsiwek:
Thanks for explaining.  I know the builtin `configparser` is available in Python 3, though I'd prefer the consistency of the current approach if it's at all possible.  Does Debian have it's own dependency/package for the backported version of `configparser` that you can use ?

If not, then how does Debian generally treat the use of Python 2 versus 3 in the context of packages?  e.g. is a user allowed to select to use Python 2 instead of 3 ?  And if so, then I don't think this patch would be enough to get `bro-pkg` working for such users, right?

hillu:
In Debian (at least 9.x "stretch" and beyond) there is only a [_python-configparser_](https://tracker.debian.org/pkg/configparser) package containing the module in a path to be used by Python 2 -- packages with modules for use with Python3 would be named _python3-$MODULE_, but there is no _python3-configparser_ package.

Using try/except constructs to import the correct modules seems to be quite normal in Python2/Python3 compatible code, e.g. for StringIO which has been [renamed or moved around](https://stackoverflow.com/questions/11914472/stringio-in-python3#18284900).

jsiwek:
Is there a reason why a python3-configparser doesn't exist?

The thing not making sense to me is that the backport of configparser is needed for python2 support and would also work for python3 users without any code changes, so why add complexity just so that python3 users instead use the built-in version of configparser if it's only going to be functionally equivalent?

I think a lot of python devs would reach a similar conclusion, especially those that read the docs of the configgparser backport -- the impression you get is that it's clearly meant to be used in both python2 and python3 and the first example of how to import it is exactly what I did here.  So maybe there is actually more that can/should be done in the Debian packaging for the configparser backport in order to support the intent of that library?

I know this patch isn't actually a terrible amount of complexity to add in just this case, but it also seems more efficient in the larger context if Debian were to have a python3-configparser package -- e.g. it's the difference between a single downstream change versus having to potentially send multiple patches/PRs to various other upstream python projects that took a similar approach to the one here.

hillu:
I can only take a guess here. From where I stand, it looks as if the Python community had been divided for a long time between those who are content Python2 users and those who are excited about improvements and other features in Python3 and no longer care that much about Python2 because "Python3 is the future". The standard library of Python3 has introduced some new modules and thrown away some. Python2 users still want to get access to some of the shiny Python3 modules, but Python2 at htis point is pretty much in maintenance mode, so specific modules were backported and put into a "backports" namespace.

As for a "python3-configparser" package in Debian -- it is not really needed because the code is already there, in Python3 the standard library, just in a different namespace.

If you dislike the "added complexity" so much that you don't want to merge, this is fine by me -- I suppose we'll just keep this patch and #8 in the Debian package until distributions start dropping support for Python2.

jsiwek:
> If you dislike the "added complexity" so much that you don't want to merge, this is fine by me -- I suppose we'll just keep this patch and #8 in the Debian package until distributions start dropping support for Python2.

It's more that I was hopeful that an ideal scenario for everyone would have looked like "all python devs can write this simpler import line and it will work for both python2 and python3 and continue working even after python2 EOL".

I'm fine merging this if you don't think there's a chance of a "python3-configparser" package being made or that it will get abandoned anyway after python2 is dropped.  I know it's not technically needed as you stated, but I'd say it's wanted because the the ideal scenario above is not possible without it.  Just let me know what you think.

SleuthKid:
The problem here is:

There is no backports package in python3 namespace. So your import fails for python3.

There is a configparser package in python3 that has to be imported from the correct configparser namespace.

For me this PR is necessary for all default python3 distributions outside the sometimes self-healing pip landscape :)

jsiwek:
> For me this PR is necessary for all default python3 distributions outside the sometimes self-healing pip landscape :)

A non-pip, python3 user is still able to manually install the configparser backport and get bro-pkg working, right?  Such a user already has the burden of manually installing other dependencies, so not a big deal to also install this one, even if it seems redundant.

Another benefit of always depending on the backport, even on python3, is that it decouples that dependency from the python version of the user -- if there's a bug in python 3.Y.Z's configparser, bro-pkg is isolated from the effects of that.  Everyone would be relying on the same configparser code base and that would hopefully be easier to patch/update than changing their python version.

i.e. bro-pkg has chosen to depend on a third-party implementation of configparser that happens to have the goal of being functionally equivalent to python3's configparser.

I know I'm making a big stink for a PR that's fairly simple, though I'm still thinking the principle of the matter suggests the ideal solution lies with Debian providing a python3-configparser package.  If you tell me that's definitely not going to happen or a major hassle to find out, I'll go ahead and merge the PR, though I'm hoping you'd still agree with my visions for ideal software dev dreamlands :)

hillu:
Most likely, there will not be a python3-configparser package.

Whether you decide to me or reject this patch and #8 -- we'll know what to do for the bro-pkg package in Debian. :-)

jsiwek:
I'll merge.  Thanks for patiently entertaining my idealisms.

SleuthKid:
With merging of #8 I will close this PR now.
Thanks Jon!

PR #29, opened 2018-03-28
Add btest as a requirement

JonZeolla:
I was running through an exercise of packaging `bro-pkg` as a deb this morning and noticed that `requirements.doc.txt` doesn't mention btest as a requirement, but `doc/quickstart.rst` does.

### DIFF
--- a/requirements.doc.txt
+++ b/requirements.doc.txt
@@ -4,3 +4,4 @@ GitPython
 semantic_version
 configparser
 sphinx_rtd_theme
+btest


jsiwek:
`requirements.doc.txt` is specifically what's required for building the documentation, not for general `bro-pkg` usage.  It's meant either for use by either developers that want to build docs locally or by automatically triggered Read the Docs rebuilds.

I imagine there could even be versions of the requirements file: "doc", "min", and "all".  Or maybe it would be better to just merge everything into a single file even if it will pull in documentation-related packages that users wouldn't typically use?  What would work best for you?  Or what's common amongst other python packages?  Open to suggestions.

JonZeolla:
Ah, okay that makes sense.  In my case I wouldn't mind any of the options you suggested, as long as there is a `requirements*.txt` that contains all of the dependencies.

jsiwek:
It's now just a `requirements.txt` w/ everything in it, thanks.

PR #33, opened 2018-06-16
fixes branch installs when version name contains a slash

dopheide-esnet:
While testing package updates for Broker, we found that specifying a version didn't work if the branch name contained a '/'.

For instance:
bro-pkg install --nodeps --version feat/broker https://github.com/ncsa/bro-is-darknet

Pulls everything down, runs the tests, then fails the install because it sees the available version name as "broker" instead of "feat/broker".  Resulting in:

Failed installing "https://github.com/ncsa/bro-is-darknet": no such branch or version tag: "feat/broker"

### DIFF
--- a/bropkg/manager.py
+++ b/bropkg/manager.py
@@ -2058,7 +2058,7 @@ class Manager(object):
                 status.tracking_method = 'branch'
 
         status.current_version = version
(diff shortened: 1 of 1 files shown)


jsiwek:
Thanks for that, in v1.3.12 now.

PR #48, opened 2019-04-03
bro-pkg.meta filename access via the Python API

ckreibich:
This tracks the absolute file name to bro-pkg.meta in PackageInfo instances, so that folks who want to parse the metadata themselves can access it reliably.


jsiwek:
Thanks, those changes now included in `v1.6.0`

PR #58, opened 2020-03-14
Support `@` in source paths to denote branch/revision to use

awelzel:
Allow an `@` character in the path part of an URL to be used as a specifier
which branch/tag/commit to checkout. Shamelessly copied from `pip` where it
can be used to specify the revision of a dependency.

The main motivation is to support testing changes to source repositories in branches
without the need to fork. In addition, the commit adding `--extra-url` enables easier testing
of different forks (and branches) without editing the config file. It also allows to override
the URL of existing sources.

I admit I only tested the changes with `zkg search` (that works), but posting it here to
see if that approach would be feasible and something that could possibly be accepted?


jsiwek:
Thanks, definitely looks like an improvement/convenience and the changes all look good and I'd be ready to merge whenever, provided you don't have further changes planned.  If you don't mind trying to extend the tests to cover the new functionality in some basic sense, that would also be a big help.  The tests can all be run locally via `btest` (or `make test`) if you have Zeek installed and in `PATH` (and as of a few minutes ago, this repo now uses Cirrus CI if you want to rebase on `master` and have the PR go through it).

awelzel:
Hey Jon, thanks for the feedback (and setting up the CI)!

I've been able to run the tests locally and will add some that exercise the added code paths.

awelzel:
@jsiwek - I've pushed some tests that should at least fail if something is completely off. I also removed the `[RFC]` part from the commit.

Let me know if there's something else open you'd like to revisit.

jsiwek:
Awesome job, thanks.  It's now in `zkg v2.1.0`.

PR #64, opened 2020-04-03
Introduce basic package name validation

awelzel:
Hey,

I tried to install a package from `/packages` today (docker volume mount, long story, ;-) ).

Anyway, it fails:
```
root@aee775ebc7e9:/packages# zkg install --force .
Running unit tests for "/packages"
error: failed to run tests for /packages: could not create symlink at /root/.zkg/testing/packages/scripts/packages: FileExistsError: [Errno 17] File exists: 'packages/packages' -> '/root/.zkg/testing/packages/scripts/packages'
```

This MR includes a test to reproduce the issue only.

Any thoughts on how to "react" here? It seems the pragmatic to just respond with a message saying `packages` is reserved? And probably add `package` as well?

Or just too esoteric to be dealt with?

### DIFF
--- a/testing/tests/install-invalid
+++ b/testing/tests/install-invalid
@@ -1,0 +1,12 @@
+# Test invalid package names
+# @TEST-EXEC: ! bash %INPUT 2>output
+# @TEST-EXEC: btest-diff output
+
+CONFIG=$(pwd)/config
+mkdir -p invalid
+
+cp -R ./packages/bar ./invalid/\ fronting-whitespace
+(cd ./invalid/\ fronting-whitespace && zkg --config=$CONFIG install --force .)
+
+cp -R ./packages/bar './invalid/trailing-whitespace '
+(cd './invalid/trailing-whitespace ' && zkg --config=$CONFIG install --force . )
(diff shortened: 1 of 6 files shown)


jsiwek:
Yeah, it is a bit esoteric, but also likely simple enough to teach `zkg` the concept of "reserved" package names, initially reserve "packages" and "package", and reject them with nice error message like you suggest if you're interested in trying to make a patch.

awelzel:
@jsiwek - I came up with something. I called it invalid package though to make it a bit wider applicable.

testing/tests/install-invalid
```diff
@@ -0,0 +1,12 @@
+# Test invalid package names
+# @TEST-EXEC: ! bash %INPUT 2>output
```
jsiwek, line 2:
```suggestion
# @TEST-EXEC-FAIL: bash %INPUT 2>output
```

Think that's a bit clearer and just going to change in the merge, but thought to mention in case you didn't know about it (or just let me know if you had other reason to prefer the original way).

jsiwek:
Now in v2.1.2, thanks.

testing/tests/install-invalid
awelzel: Perfect, wasn't aware this existed - thanks for letting me know!

PR #65, opened 2020-04-06
Accept Y as yes, too.

awelzel:
I've just ran into this. I should've just hit enter, or typed `y`, but
aborting on `Y` was very unexpected (apt-get allows `Y`).

    $ zkg --config ./config bundle --manifest ./manifest -- bundle.tar.gz
    ...
    Proceed? [Y/n] Y
    Abort.


awelzel:
Sorry for the noise - just saw we read with `.lower()`.

PR #69, opened 2020-09-21
Add `package_base` as a pre-defined configuration key for value interpolation

rsmmr:
This enables a package to gain access to files from another package in
a well-defined manner. For example, if package `foo` wants to access
file `xyz.cfg` from package "bar", it can now declate `bar` as a
dependency and then use `%(package_base)s/bar/xyz.cfg` from its
`build_command`.


jsiwek:
zkg v2.3.0 is now out with this feature.

PR #72, opened 2020-09-25
When running tests, use any already installed dependencies rather than rebuilding them.

### LINKED ISSUE #70: Unnecessary dependency rebuilds
I'm confused by something I'm seeing and not sure if that's intended or not:

Assume we have a package `A` that depends on another package `B`. I install `B` manually first: `zkg install B`. Then I do `zkg install A`. It seems that during the unit testing phase of `A`,  `B` gets cloned and built again; l see both packages in `testing/A/clones/{A,B}` and I also saw errors being reported from that build of `B` (the errors were unrelated to this issue, but it shows that the build was happening). I find that surprising and potentially dangerous, as I would expect `A` to test against the already installed version of `B`.

rsmmr:
Closes #70.

@jsiwek if you're fine with this change and merge it, please also push another release out to pypi.

### DIFF
--- a/zeekpkg/manager.py
+++ b/zeekpkg/manager.py
@@ -2097,7 +2097,7 @@ class Manager(object):
         make_dir(stage_plugin_dir)
 
         request = [(package.qualified_name(), version)]
-        invalid_deps, new_pkgs = self.validate_dependencies(request, True)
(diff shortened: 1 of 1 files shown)


jsiwek:
Patch is in v2.3.1

PR #87, opened 2020-12-01
Detect default branch name of packages automatically

### LINKED ISSUE #76: Allow non-default branch names
This is kind of a continuation of #74.

While the specific problem for #74 is fixed, I also think that we should give people a way to use non-master branches as the default branch in their repository. As mentioned before - there is an increasing push to not use master as the default name in git - and I think that it would be nice to support this.

jsiwek:
Fixes #76 

Could use a sanity check/review to see if I missed anything important like:
* Is there actually some concept of a "default" branch that can be used in absence of a remote HEAD reference ?
* Whether change in default branch selection logic disrupts any particular use-case in a notable/likely way (my guess is 'no'): read commit message of https://github.com/zeek/package-manager/commit/99a3951ec8f74579ef0e0e04f2bb37d617b31918 for exactly what the logic now looks like

### DIFF
--- a/doc/package.rst
+++ b/doc/package.rst
@@ -755,11 +755,11 @@ remote repository:
 
 Alternatively, if you expect to have a simple development process for
 your package, you may choose to not create any version tags and just
-always make commits directly to your package's *master* branch.  Users
-will receive package updates differently depending on whether you decide
-to use release version tags or not.  See the :ref:`package upgrade
-process <package-upgrade-process>` documentation for more details on
-the differences.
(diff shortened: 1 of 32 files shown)


0xxon (approved):
Thanks a lot, this looks great.

The only oversight that I found might be the documentation of --version in zkg.1 - that currently refers only to master, and I think might now also take main or other branch names.

jsiwek:
Cool thanks (the zkg.1 man page will get autogenerated from other source code / docstrings).

PR #90, opened 2021-01-05
Add the Zeek bin directory to PATH by default

ckreibich:
This helps package installations succeed that require executables from that directory, such as zeek-config. @0xxon noticed this in her package-making adventures. For testing, I removed my installation's bin directory from my PATH and a plugin-package needing `zeek-config` installed fine with the fix (and indeed doesn't otherwise).

### DIFF
--- a/zkg
+++ b/zkg
@@ -44,7 +44,7 @@ ZEEK_BIN_DIR = '@ZEEK_BIN_DIR@'
 if os.path.isdir(ZEEK_BIN_DIR):
     try:
         if ZEEK_BIN_DIR not in os.environ['PATH'].split(os.pathsep):
-            os.environ['PATH'] += os.pathsep + ZEEK_BIN_DIR
+            os.environ['PATH'] = ZEEK_BIN_DIR + os.pathsep + os.environ['PATH']
     except KeyError:
         os.environ['PATH'] = ZEEK_BIN_DIR
 else:
--- a/zkg
+++ b/zkg
@@ -37,6 +37,19 @@ if os.path.isdir(ZEEK_PYTHON_DIR):
 else:
     ZEEK_PYTHON_DIR = None
 
+# Similarly, make Zeek's binary installation path available by
+# default. This helps package installations succeed that require
+# e.g. zeek-config for their build process.
+ZEEK_BIN_DIR = '@ZEEK_BIN_DIR@'
+if os.path.isdir(ZEEK_BIN_DIR):
+    try:
+        if ZEEK_BIN_DIR not in os.environ['PATH']:
+            os.environ['PATH'] += os.pathsep + ZEEK_BIN_DIR
+    except KeyError:
+        os.environ['PATH'] = ZEEK_BIN_DIR
+else:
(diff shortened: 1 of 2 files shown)


0xxon:
Thanks a lot :)

This would be nice to have in 4.0

zkg
```diff
@@ -37,6 +37,19 @@ if os.path.isdir(ZEEK_PYTHON_DIR):
 else:
     ZEEK_PYTHON_DIR = None
 
+# Similarly, make Zeek's binary installation path available by
+# default. This helps package installations succeed that require
+# e.g. zeek-config for their build process.
+ZEEK_BIN_DIR = '@ZEEK_BIN_DIR@'
+if os.path.isdir(ZEEK_BIN_DIR):
+    try:
+        if ZEEK_BIN_DIR not in os.environ['PATH']:
```
jsiwek, line 46: That `in` operator would only do a substring-search which may not be correct.

```suggestion
        if ZEEK_BIN_DIR not in os.environ['PATH'].split(os.pathsep):
```

zkg
```diff
@@ -37,6 +37,19 @@ if os.path.isdir(ZEEK_PYTHON_DIR):
...
+            os.environ['PATH'] += os.pathsep + ZEEK_BIN_DIR
```
jsiwek, line 47: Would prepending better guarantee matching against the associated Zeek install ?

```suggestion
            os.environ['PATH'] = ZEEK_BIN_DIR + os.pathsep + os.environ['PATH']
```

jsiwek (changes requested):
@ckreibich thanks, see if my suggestions are sensible

@0xxon no problem putting it in 4.0, just curious/checking I understand the usage issue: is it only that the Zeek `$prefix/bin/` isn't in your `PATH` and you want to invoke `zkg` via absolute path (without any `PATH` changes) ?

zkg
ckreibich: Right, that's better. Thanks!

zkg
ckreibich: Yeah I thought about that one too and wasn't sure. Johanna's problem was just that `zeek-config` wasn't found at all. Looking at it, I like it. It makes things more predictable. Thanks!

0xxon:
> @0xxon no problem putting it in 4.0, just curious/checking I understand the usage issue: is it only that the Zeek `$prefix/bin/` isn't in your `PATH` and you want to invoke `zkg` via absolute path (without any `PATH` changes) ?

Yup.

This was, basically, something that I noticed while putting zkg into our binary packages. Due to the fact that the binary packages are installed in /opt/zeek, and not added to the ``PATH`` by default, it is not completely unlikely that people will just start calling executables using their full path - without them being available in ``PATH`` - I did that for testing.

It just seemed unintuitive for me that zkg would not be able to find the Zeek installation that it is part of, without the ``PATH`` entry.

jsiwek:
Patch is now in zkg v2.6.1 and used by zeek `master` and `release/4.0` branches.

PR #91, opened 2021-01-21
Add `--user` mode in zkg

ckreibich:
This is a draft for now as it's not yet complete ... I'd like to get a sense to which this looks right to you @jsiwek. I'll look into test cases then. I've verified that if I make my Zeek installation read-only, the bundled zkg is able to install and list packages fine with `zkg --user`.

I've not yet considered transitional issues when users have pre-existing `~/.zkg/config` files. I considered just completely ignoring an existing config file in `~/.zkg/config` when using `--user`, but I'm not sure that's a good idea — it would preclude the use of other, legitimately adjustable config settings in that file in the future.

The tweak to `zkg env` is just a detail — I'd noticed that with my existing environment variables it actually ended up listing paths redundantly, so I figured we might as well avoid that.


jsiwek:
> I've not yet considered transitional issues when users have pre-existing `~/.zkg/config` files. I considered just completely ignoring an existing config file in `~/.zkg/config` when using `--user`, but I'm not sure that's a good idea — it would preclude the use of other, legitimately adjustable config settings in that file in the future.

Think it would work fine to not even let `--user` interfere with config-file-locating logic, just when it comes time to determine the "final" value of certain settings (currently `state_dir`, `script_dir`, or `plugin_dir`), have `--user` override everything ?

Otherwise, the changes are just like what/where I'd expect, thanks for helping with this.  A polishing/finalization wishlist if you'd like to take a look at any:

* Add test cases
* Error message when encountering permission issues might be changed to suggest `--user`
* Documentation:
  * Call out the overriding capability of `--user` for the settings it impacts in https://docs.zeek.org/projects/package-manager/en/stable/zkg.html#config-file
  * Maybe another place to mention `--user` is in the blue "Note" box here since it's already describing the pitfall of filesystem permissions: https://docs.zeek.org/projects/package-manager/en/stable/quickstart.html#basic-configuration

ckreibich:
> Think it would work fine to not even let --user interfere with config-file-locating logic, just when it comes time to determine the "final" value of certain settings (currently state_dir, script_dir, or plugin_dir), have --user override everything ?

Yeah ... I was actually wondering about that too. I see pros and cons to either approach:

* With the approach implemented, there's an implicit statement that `--user` will lead to the use of `~/.zkg/config`, or the built-in defaults when that file doesn't exist. So when somebody goes in and edits `~/.zkg/config` to point `script_dir` etc at something else, `--user` might still end up touching stuff outside `~/.zkg`, though that'd then be the user's own doing.
* With your approach it could be confusing that you end up using one config file but the settings in effect are others — the users won't necessarily know about the meaning of `script_dir` etc. So the user might then end up saying `--configfile ~/.zkg/config --user`, which seems a bit odd. But I like that this approach clearly overrides any existing values, so the transition to the new approach is clear: it applies the moment you say `--user`.

I landed on the former, but don't feel strongly here. For both approaches, I like the idea that `zkg --user autoconfig` would produce a `~/.zkg/config` that has all settings point inside `~/.zkg` and that's a proper config file (i.e. can contain other settings) going forward.

ckreibich:
Oh and yes, agree to the remaining todo list! I'll get on those.

jsiwek:
> I landed on the former, but don't feel strongly here.

Me neither since either Works as long as their behavior is documented, so I'd stick with what you have already.  Maybe additionally clarify the "help" description of `--user` to say it's more about forcing use of `$HOME/.zkg/config` if it exists or else use of default config settings that will all point within `$HOME/.zkg/`

ckreibich:
Cool, that all sounds great to me. Thanks Jon. I should be able to wrap this up tomorrow am ... I'm a bit behind tonight due to an unforeseen childcare window.

ckreibich:
Testing this proved a lot more time-consuming than I anticipated. I'll need to dig in to see what Cirrus dislikes. Will resume in a bit.

ckreibich:
Okay Jon, let me know your thoughts. I initially had a more comprehensive btest that included the `--user` suggestion in a baseline, but this proved tricky when running the tests as root, because root tends not to have problems writing where it likes. :) In the end I dropped this part from the test, which I think is okay. If you'd like to change things, let me know or feel free to just go ahead.

I noticed that the zkg test baselines don't yet have the new btest header line? I hit one CI failure where I had introduced that header through a refresh on my side — not sure why that triggered an error. I omitted the header in the new baseline for now.

jsiwek:
@ckreibich thanks

> I noticed that the zkg test baselines don't yet have the new btest header line? I hit one CI failure where I had introduced that

Maybe due to the Docker images being created a long time ago and the btest version it uses is whatever it was at that time (it's not installed/updated every run).  The image for CI will at least get updated when Zeek 4.0.0 is released, so can also look at updating baselines/btest at that time.

PR #100, opened 2021-03-18
Add support for packages shipping executables.

### LINKED ISSUE #97: Support installing executables with a package
Right now zkg doesn't have any support for installing additional, non-Zeek pieces that come with a package. An example I ran into with https://github.com/zeek/spicy-plugin is an additional binary: the package compiles a tool `spicyz`. That currently ends up inside a deeply hidden directory of the zkg clone tree. Hard to tell users to run it from there. 

Idea: enable a package to specify executables in `zkg.meta`, e.g.:

```
executables = build/bin/spicyz
```

zkg could then create a managed directory `.zkg/bin` and maintain symlinks pointing from that directory to the actual binaries. The symlinks would get in place when a package is installed, and be removed when it's uninstalled. Users would then just once have to put `.zkg/bin` into their PATH and have it working for all packages.

I'm not sure if there could be more general mechanism going beyond just executables, but can't think of anything right now.

rsmmr:
This introduces a new "executables" metadata field that a package can
set to scripts or binaries that it wants to make available to users. On
package installation, zkg will link these executables into a new,
central "bin_dir" that users can put into their PATH. By default, that's
"<state_dir>/bin", but it can be relocated through a corresponding "bin_dir"
entry in the configuration file. On package removal, these symlinks get
removed from "bin_dir". `zkg env` now updates the PATH to include "bin_dir", and users can also do that manually 
by setting `PATH=$(zkg config bin_dir):$PATH.`

Implements #97.


rsmmr:
Added some docs, and a couple further code tweaks.

zeekpkg/manager.py
```diff
@@ -2239,6 +2283,16 @@ def _stage(self, package, version, clone,
 
         raw_metadata = _get_package_metadata(raw_metadata_parser)
 
+        # Ensure any listed executables exist as advertised. Do this first so
+        # that we don't build anything if it fails.
```
jsiwek, line 2287: Don't understand the reason for this check happening before execution of `build_command`.  Can that not end up producing the desired executable to install ?  The new docs showed an example with `build/a.out` which I took to mean as implied/intended support for `build_command` producing the executable.

jsiwek (commented):
Cool, I already made the couple changes I commented on inline except for the one about how this is meant to interact with `build_command`.

Also updated the zeek-bundled config file: https://github.com/zeek/zeek/commit/c553df5dfe50d530dbf0da1cb79cf8db0660e0d6

zeekpkg/manager.py
rsmmr: Oh, you're absolutely right. That ordering doesn't make sense, the check needs to move after the build. I'll do a quick check and then another PR later.

rsmmr:
Thanks Jon!

PR #102, opened 2021-03-24
Templating support for zkg

### LINKED ISSUE #93: Templating support for zkg
I've put together a prototypical version of package instantiation/templating support in `zkg` and welcome your feedback. The focus here is on the overall model, which is as follows:

- Templates are independent git repositories, much like Zeek packages. This brings all the usual support for versioning via branches and tags.
- `zkg` knows of one such repo by default, similar to its default package source, but you can provide others at the command line. (We can consider shipping a template by default with Zeek in case there's value in having a template to start from without network connectivity, but that is not a top concern.)
- Templates provide an installable, running Zeek package in form of a file tree that gets instantiated via `zkg`.
- Templates may provide optional "features" that get overlaid onto this package, such as plugins, Spicy support, CI scripting, documentation, additional testing, etc.
- Templates also hook directly into `zkg` via Python, a bit like plugins. This means templates can run code in order to instantiate or modernize existing content, with all of the power that this entails. The observation here is that a plain text patching model quickly becomes insufficient in order to tackle code modernization or surgical file updates, e.g. to an existing zkg.meta. The assumption here is that we trust templates as input, but since these generate Zeek packages, plugins, etc, that's a given.
- The primary audience for template _development_ is us, while the primary audience for _applying_ the templates is any `zkg` user. The design enables others to provide or use alternative templates, but the main goal here is to make flexible package bootstrapping available to any `zkg` user while letting us provide really powerful templates.

A bit of discussion:
- Why not Cookiecutter? Cookiecutter is cool and we could use it, but it'd only be a building block. Our requirements are a bit different: we want to integrate tightly into `zkg`, allow the user to combine features into the output as needed, and our actual text templating needs are really basic.
- Why not Cruft? I've played around with Cruft and while its premise (that you can update generated content via patching when the underlying template gets updated) is really appealing, in our case I don't think it holds up well in practice. Cruft's premise is strong when the generated content doesn't get modified, but that's generally not true for us. When you do modify the generated content, there comes a point when Cruft can no longer apply its patches. That itself could be fine, but it gets a bit bumpy:
```
$ cruft update
Respond with "s" to intentionally skip the update while marking your project as up-to-date or respond with "v" to view the changes that will be applied.
Apply diff and update? (y, n, s, v) [y]: v
diff --git 1/tmp/tmpq1tfg3jw/current_template/README.rst 2/tmp/tmpq1tfg3jw/new_template/README.rst
index 19a29e6..d00a0d2 100644
--- 1/tmp/tmpq1tfg3jw/current_template/README.rst
+++ 2/tmp/tmpq1tfg3jw/new_template/README.rst
@@ -26,7 +26,7 @@ If you already have Zeek and ``zkg`` installed, simply run:

 .. code-block:: console

-    zkg install https://github.com/ckreibich/foobar
+    zkg install --force https://github.com/ckreibich/foobar

 If this is being installed on a cluster, install the package on the manager, then deploy it via:

Respond with "s" to intentionally skip the update while marking your project as up-to-date or respond with "v" to view the changes that will be applied.
Apply diff and update? (y, n, s, v) [y]:
error: git diff header lacks filename information when removing 1 leading pathname component (line 5)

Failed to apply the update. Retrying again with a different update stratergy.
error: git diff header lacks filename information when removing 1 leading pathname component (line 5)

Project directory may have *.rej files reflecting merge conflicts with the update. Please resolve those conflicts manually.
Good work! Project's cruft has been updated and is as clean as possible!
```
The modification I made to the generated content was `zkg install` -> `zkg install --user`, which the output doesn't tell you. Since the underlying approach is really simple (instantiate former template, instantiate new template, diff the two, apply to current content) I think we're better off implementing this ourselves, allowing also programmatic updates (that could be a regex replace for a renamed symbol, a `configparser` update to zkg.meta, etc). Finally, the dependencies pulled in by the two are pretty significant, on the order of 20 additional packages total. One can definitely argue here that it doesn't matter since the dependencies install easily, and they do. But it just looks like overkill to me.

A bit of detail about the current prototype. It comes in at around 300 lines total, so it's not that much. I've put together a simple template that provides a running package with a passing btest, an optional additional plugin, likewise with a test, and a really simple Github CI setup. This looks as follows:
```
$ zkg create --help
usage: zkg create [-h] --packagedir DIR [--force] [--version VERSION]
                  [--features {github-ci,plugin} [{github-ci,plugin} ...]]
                  [--template URL] --name STRING [--namespace STRING]

This command creates a new Zeek package from template "package-template" in
the directory provided via --packagedir. If this directory exists zkg will not
modify it. Use --force to allow zkg to delete and recreate it.

optional arguments:
  -h, --help            show this help message and exit
  --packagedir DIR      Output directory into which to produce the new
                        package. (default: None)
  --force               If the package output directory exists, remove and re-
                        create it. (default: False)
  --version VERSION     The version of the package to install. A version tag,
                        branch name, or commit hash may be specified here. If
                        --template refers to a local git repo with a working
                        tree, then its currently active branch is used. The
                        default for other cases is to use the latest version
                        tag, or if a template has none, the default branch,
                        like "main" or "master". (default: None)
  --features {github-ci,plugin} [{github-ci,plugin} ...]
                        Additional features to include in your package.
                        (default: None)
  --template URL        By default, zkg uses its own package template. Use
                        this to select an alternative one. (default: None)
  --name STRING         The name of the package, e.g. "FooBar". Required.
                        (default: None)
  --namespace STRING    A namespace for the package, e.g. "MyOrg". Required
                        only when including a plugin in your package.
                        (default: None)
```
Everything template-specific (the features, the string values you can provide) come from the template and aren't coded into `zkg`. It's actually a new clone in `zkg`'s internal state:
```
$ cd ~/.zkg/clones/
$ ls -la
total 20
drwxr-xr-x. 5 christian christian 4096 Feb 22 15:16 ./
drwxr-xr-x. 7 christian christian 4096 Feb 22 15:16 ../
drwxr-xr-x. 2 christian christian 4096 Feb 22 15:16 package/
drwxr-xr-x. 3 christian christian 4096 Feb 22 15:16 source/
drwxr-xr-x. 3 christian christian 4096 Feb 22 20:01 template/
$ cd template
$ ls -la
total 12
drwxr-xr-x. 3 christian christian 4096 Feb 22 20:01 ./
drwxr-xr-x. 5 christian christian 4096 Feb 22 15:16 ../
drwxr-xr-x. 7 christian christian 4096 Feb 22 20:03 package-template/
$ cd package-template/
christian@hogthrob4:(master)…ate/package-template> git remote -v
origin  https://github.com/ckreibich/package-template (fetch)
origin  https://github.com/ckreibich/package-template (push)
```

To create a package with a plugin and Github CI, you might say:

```
$ zkg create --name Foobar --namespace MyOrg --feature plugin github-ci --packagedir foobar
$ cd foobar
$ git log -1
commit b7384fda1ce8b1a6aa725b7f896d506f9ecf791c (HEAD -> master)
Author: Christian Kreibich <christian@corelight.com>
Date:   Mon Feb 22 20:15:13 2021 -0800

    Initial commit.

    zkg 2.7.1 created this content from template "package-template"
    using version master with features "github-ci" and "plugin".
$ cat zkg.meta
[package]
script_dir = scripts
build_command = ./configure && cd build && make
test_command = cd testing && btest -c btest.cfg
summary = TODO: A summary of Foobar in one line
description = TODO: A more detailed description of Foobar.
        It can span multiple lines, with this indentation.

[template]
source = package-template
version = master
zkg_version = 2.7.1
features = github-ci,plugin
arg.name = Foobar
arg.ns = MyOrg
$ zkg install .
  /home/christian/t4/foobar (master)

Proceed? [Y/n]
Running unit tests for "/home/christian/t4/foobar"
Installing "/home/christian/t4/foobar"......
Installed "/home/christian/t4/foobar" (master)
Loaded "/home/christian/t4/foobar"
```
To use an alternative template, you'd say `zkg create ... --template https://github.com/some/other-template` etc. `zkg'`s help output will reflect the supported features and arguments for that template.

The current CI feature is Github-only and very simple, thanks to @0xxon's Zeek RPMs. It's a POC, not meant for future use. @grigorescu's [cookiecutter-zeekpackage](https://github.com/esnet/cookiecutter-zeekpackage) has a ton of cool stuff particularly for CI, and that's a separate thing to iterate on in the future. That said, you can see the current CI running in an instantiated package [here](https://github.com/ckreibich/testpackage/actions). It has a per-commit and a nightly workflow, using straight-up scripting on the Github VMs (again, this is POC). Attentive readers will notice that with plugins the Zeek LTS (3.0.x) test bombs — that's because the plugin feature currently doesn't provide compatibility for the 3.0.x series!

If you'd actually like to give this a go, you can grab my branch and the package template that it uses by default below. I've not yet cleaned up the commits and the code will need an edit pass.
- https://github.com/ckreibich/package-manager/tree/topic/christian/package-templates-decoupled (the meat here is in [template.py](https://github.com/ckreibich/package-manager/blob/topic/christian/package-templates-decoupled/zeekpkg/template.py)).
- https://github.com/ckreibich/package-template

Again, the focus here isn't on what the template does right now, but the overall approach. What do you guys think? Basically cool? Basically not cool?

ckreibich:
This is a more fully baked version of the templating support we've previously discussed over in #93. The idea is as follows:

- `zkg` gains support for instantiating new Zeek packages from a package template.
- These templates are independent git repositories, with corresponding support for versioning, branches, etc.
- Templates provide two kinds of functionality: Zeek packages and optional additional features, such as CI, plugins, Spicy support, specific copyrights, etc. You can mix-and-match these into your resulting package.
- Templates support named parameters that get substituted upon instantiation. The user provides these parameters as user vars (via user prompting, environment variables, previously stored values in the zkg config, or `--user-var` -- more on this below). The template expands the given user variables into its required parameters. This is an n:m relationship -- for example, given a package name, the template might define one parameter with that name as given, one sanitized and upper-cased for use in a module string, one derived from a combination of user vars, etc. User vars are the external interface, parameters their internal application.
- Templates are mainly a tree of file content, but driven by Python code (`__init__.py` at the toplevel). This enables flexibility with input validation, how to instantiate content, deal with potential conflicts across features (overwriting vs selectively editing files like zkg.meta, for example), or modernizing content (say via patching, moving files around, renaming things, overwriting them, sanity-checking existing content, etc).
- We maintain an official default template at https://github.com/zeek/package-template, but users can define and use their own.
- The target audience for template development is primarily us Zeek devs, the audience for template use is anyone who builds Zeek packages.

What does this address?

- A more accessible package creation mechanism. `init-plugin` provides a solid layout but resides only in the source tree.
- A more flexible package creation mechanism: pick just the package features you need, with many at your disposal.
- A more explicit default source of truth for what Zeek packages should look like.
- Assistance (with caveats) with keeping package content modern over time.

What is not there yet?

- An agreed-upon v1 template. https://github.com/zeek/package-template exists and works (providing plugins and really basic Github CI as starting points), but people will have thoughts/opinions about what should go in there. I suggest we label this templating feature beta to allow ourselves time to iterate on the template. I'm definitely open to suggestions here.

- Modernization of an existing template-derived package. This is a natural next step but the PR is already on the large side, so I figured I'd wait. (I think such modernizations will likely end up being a mix of doable approaches when a template clearly owns content, and suggestions/guidance for modernization when things are less clear, e.g. via conflicting patches. Regardless, I do think they will be useful for the user.)

- A command that prints information about a package (see below). This should be there because it's otherwise difficult for the user to know what features a package supports, what versions it offers, etc. This is much like `zkg info` for packages.

- Support for combining multiple templates. For example, one might want to use our default template, but overlay a CI feature for a specific platform. I feel this would often be valuable, and simpler to maintain than forking the standard one and maintaining additional features on top of it.

Other things in this PR:

- I added two things to zkg's user-var support: you can now enter user vars via the command line, using `--user-var NAME=VAL`. This works for any zkg command that currently prompts for user-var input. I've also tweaked the way user input works: the default or pre-recorded value of a user variable now gets populated into the input prompt. This maintains the previous behavior of letting you "just hit enter" to confirm the existing value, but also lets you edit the value, including setting the variable to an empty string.

- I noticed that on my systems the btest output got cluttered with git warnings about the lack of a configured default branch name (`init.defaultBranch`), which newer git versions request. I added this to the `scripts/git` wrapper.

If you'd like to give this a go:

- Install this zkg as usual, then try e.g. `zkg create --packagedir foobar` or `zkg create --packagedir foobar --feature plugin github-ci --user-var name=FooBar --user-var namespace=MyOrg`.

You can clone https://github.com/zeek/package-template to your system and point `zkg create` at it via the --template argument if you'd like to tweak the template.

Things I'm unsure about:

- Command naming. "zkg create" is concise, but it seems likely we'll want a `zkg template` command set, e.g. to have `zkg template info`, `zkg template update`, etc. So template instantiation could live there from the outset, perhaps as `zkg template build`.

- What happens to `zeek-aux/plugin-support`? My suggestion would be to reduce it to what its name suggests: plugin support. We want to continue standalone plugins, for which this seems a good fit.

Sorry for the size of the whole thing. :grimacing: 

Resolves #93.


ckreibich:
I realized I'd like to make a tweak to allow user vars to be tied to specific features instead of the template as a whole — that way, when a feature requires user input that input will only happen when template instantiation actually uses that feature.

Also, there's no documentation yet. I'll be happy to work on that as this moves forward.

zeekpkg/package.py
```diff
@@ -85,42 +86,6 @@ def short_description(metadata_dict):
     return rval.lstrip()
 
 
-def user_vars(metadata_dict):
```
jsiwek, line 88: Don't know how likely it is someone relies on this API, but since it's not originally marked as internal in any way, I'd just leave it be an alias for `UserVar.parse_dict`.

zeekpkg/manager.py
```diff
@@ -2289,7 +2292,18 @@ def _stage(self, package, version, clone,
...
+
+        # Ensure any listed executables exist as advertised. Do this first so
+        # that we don't build anything if it fails.
+        for p in self._get_executables(raw_metadata):
+            full_path = os.path.join(clone.working_dir, p)
+            if not os.path.isfile(full_path):
+                return str.format("executable '{}' is missing", p)
+
+            if not os.access(full_path, os.X_OK):
+                return str.format("file '{}' is not executable", p)
```
jsiwek, line 2304:
```suggestion
```
Guessing this sneaked in while resolving a Git conflict?  (This logic got moved to the end and shouldn't be here anymore).

jsiwek (commented):
Looks nice, thanks.  I didn't review or test things comprehensively, but I didn't have any feedback about the basic structure of it.  Not sure if you want to take that a sign to just move forward with anything else you have planned (e.g. documentation) or wait for other feedback.

> * Command naming. "zkg create" is concise, but it seems likely we'll want a `zkg template` command set, e.g. to have `zkg template info`, `zkg template update`, etc. So template instantiation could live there from the outset, perhaps as `zkg template build`.

May not be that critical -- if there's later a family of `zkg template <operation>`, seems like a `zkg create` that gets added now can turn into simple alias/shorthand for something living under `zkg template <operation>`.

> * What happens to `zeek-aux/plugin-support`? My suggestion would be to reduce it to what its name suggests: plugin support. We want to continue standalone plugins, for which this seems a good fit.

Not sure much has to be done with it except maybe try to encourage/advertise this method of package creation in docs and de-emphasize any `init-plugin`, but otherwise `init-plugin` still seems like a way to create a valid plugin template and could be left as is if people want to use it (e.g. if they're less `zkg` or "package" oriented and more into just making a standalone "plugin" for their own use).

zeekpkg/package.py
ckreibich: Whoops! Yes, thanks for catching this.

ckreibich: Agreed, yeah. Will do.

zeekpkg/manager.py
ckreibich: Whoops! Yes, thanks for catching this. I've committed your suggestion.

ckreibich:
Thanks for looking it over, Jon. I'll write up documentation and put a bit of work into the default template, so it's in decent shape when we merge this. I'll keep rebasing this branch in the meantime as needed. I'm thinking we can merge this next week then, if that sounds good to you.

ckreibich:
@jsiwek the above push makes minor changes to what you already reviewed:

- It rearranges user variables a bit inside the template content classes. I realized that previously the template couldn't share the value of a single variable across its content, which I imagine will be common (e.g. your name might be used by multiple features).
- There's a first `template` sub-command: `zkg template info` summarizes a template's properties, much like `zkg info` does for packages. I've kept `zkg create` for now since it's concise, but we could also move it over to `zkg template` now.
- Added some resilience to incorrect input for `zkg create`, pylinting, and added more helpful error/warning messages.
- Documentation. I've not yet removed the old package production walkthroughs, just put the templating docs first.

I'm thinking this is a good time to merge, but happy to hear your thoughts ... thanks!

jsiwek:
Thanks, just wondering what the thought is on how long to expect this to be treated as a "beta" feature (and e.g. `API_VERSION` of 1.0.0-rc1).  Think a zkg v2.9 release waits on that?

ckreibich:
Thanks Jon! Yes exactly — I was thinking that when we release a new zkg version the beta status can go, and the API moves to 1.0.0. I've been meaning to ask you how you version zkg (what would be 2.8.1 vs 2.9, or even 3.0).

The beta and rc are just there because I'm nervous about the fact that mostly I've been playing with this and also developing it, so I think I'm prone to missing things. If people don't report major problems in the next 1-2 weeks, I think these could go.

jsiwek:
> I've been meaning to ask you how you version zkg (what would be 2.8.1 vs 2.9, or even 3.0).

Probably what you expect:
* Major-version: significant changes, new way of doing/structuring things, or backwards incompatible breakages
* Minor-version: added new features
* Patch-level: only bug fixes

> If people don't report major problems in the next 1-2 weeks, I think these could go.

Sounds fine.  I can check back in 2 weeks and do a v2.9 if there's no concerns.

PR #108, opened 2021-06-08
Add details about the resulting changes when running `zkg refresh` with aggregation/push

ckreibich:
zkg's log output currently does not indicate whether metadata actually changed when running `refresh` with `--aggregate`, `--push`, or (most likely) both. This PR adds a bit more detail, depending on the log level. Context for this is that when running the package source refresh in CI it's nice to see this level of detail in the run's logs.

Before:
```
2021-06-07 22:51:27 DEBUG    aggegrate file in source unchanged, restore local one
Refresh package source: zeek
	No changes
	Metadata aggregated
	Pushed aggregated metadata
Refresh installed packages
	No new outdated packages
```
(Note, it says "Metadata aggregated" one way or another.)

After, at log level `DEBUG`:
```
2021-06-07 16:19:34 DEBUG    aggegrate file in source unchanged, restore local one
2021-06-07 16:22:15 DEBUG    metadata refresh: 1 additions (corelight/pingback), 3 changes (0xxon/zeek-tls-log-alternative, zeek/spicy-analyzers, zeek/spicy-plugin), 0 removals
2021-06-07 16:22:15 INFO     committed package source "zeek" metadata update
	No membership changes
	Metadata aggregated
Refresh installed packages
	No new outdated packages
```
It also expands `test.refresh` to cover `--push`, which was so far untested.


zeekpkg/manager.py
```diff
@@ -920,17 +931,44 @@ def _refresh_source(self, name, aggregate=False, push=False):
                     parser.set(qualified_name, 'url', url)
                     parser.set(qualified_name, 'version', version)
 
+                    if qualified_name not in prev_packages:
```
rsmmr, line 934: The `qualified_name` has the `index_dir` prepended, does that match what's in the `prev_packages` set?

testing/tests/refresh
```diff
@@ -8,8 +8,9 @@
 # @TEST-EXEC: zkg list all > list.out
 # @TEST-EXEC: btest-diff list.out
 
-# @TEST-EXEC: zkg refresh --aggregate >agg.out
+# @TEST-EXEC: zkg -vvv refresh --aggregate --push>agg.out 2>agg.errout
```
rsmmr, line 11:
```suggestion
# @TEST-EXEC: zkg -vvv refresh --aggregate --push >agg.out 2>agg.errout
```

rsmmr (approved):
lgtm, just one question.

zeekpkg/manager.py
ckreibich: Yes, I think so. The `index_dir` is the `corelight` in something like `corelight/zeek-quic`, computed from the local paths on disk. The logic that the previous code uses to populate the sections in the config parser hasn't changed, and `prev_packages` is just a set of those section titles. It's definitely tricky, though ... perhaps we should rename `prev_packages` to `prev_qualified_names`?

testing/tests/refresh
ckreibich: Thanks, yep! Applied.

zeekpkg/manager.py
rsmmr: Got it, it was the naming of `index_dir` that threw me off. The way it's named and computed didn't let me think we'd arrive at just that namespace label.

All good then, ready to merge from my end.

PR #110, opened 2021-06-10
Harden "@" interpretation in package source URLs

### LINKED ISSUE #107: zkg does not support user names in package source git URLs
When using a package source URL that includes a user specification (e.g. "`git@github.com:user/repo.git`") `zkg` currently treats the @ character as the delimiter for the version to check out ("`https://github.com/foo/bar@mybranch`"):

```
2021-06-07 16:30:51 WARNING  failed to clone git repo: Cmd('git') failed due to: exit code(128)
  cmdline: git clone --depth=1 --recursive git /home/christian/.zkg/clones/source/zeek --no-single-branch
  stderr: 'fatal: repository 'git' does not exist'
warning: skipped using package source named "zeek": failed to clone git repo
```

Trying to URL-encode the user fails too, but for a different reason:

```
Traceback (most recent call last):
  File "/home/christian/inst/bin/zkg", line 2702, in <module>
    main()
  File "/home/christian/inst/bin/zkg", line 2696, in main
    config = create_config(args, configfile)
  File "/home/christian/inst/bin/zkg", line 289, in create_config
    expand_config_values(config, 'sources')
  File "/home/christian/inst/bin/zkg", line 287, in expand_config_values
    config.set(section, key, value)
  File "/usr/lib64/python3.9/configparser.py", line 1201, in set
    super().set(section, option, value)
  File "/usr/lib64/python3.9/configparser.py", line 894, in set
    value = self._interpolation.before_set(self, section, option,
  File "/usr/lib64/python3.9/configparser.py", line 402, in before_set
    raise ValueError("invalid interpolation syntax in %r at "
ValueError: invalid interpolation syntax in 'git%40github.com:ckreibich/packages' at position 3
```

One possible fix is to make `zkg` recognize the purpose of the @ character — the user prefix and version suffix are pretty easy to distinguish.

ckreibich:
Python's urlparse parses SSH git URLs ("git@github.com/foo/bar") crudely, exposing us to a risk of confusing "user@" prefixes with "@version" suffixes. In those situations we now require the part before the "@" to contain no slashes.

This isn't perfect, since this stuff can get very ambiguous if you push it — is "foo@bar" user foo at host bar, or version bar at host foo? I think this is an adequate for likely hosting scenarios.

Resolves #107.

### DIFF
--- a/zeekpkg/manager.py
+++ b/zeekpkg/manager.py
@@ -443,6 +443,19 @@ class Manager(object):
         version = None
         parse_result = urlparse(git_url)
 
+        # Prepend 'ssh://' and replace the first ':' with '/' if git_url
+        # looks like a scp-like URL, e.g. git@github.com:user/repo.git.
+        # urlparse will otherwise parse everything into path and the @
+        # is confusing the versioning logic. Note that per the git-clone
+        # docs git recognizes scp-style URLs only when there are no slashes
+        # before the first colon.
+        colonidx, slashidx = git_url.find(':'), git_url.find('/')
+
+        if '://' not in git_url and colonidx > 0 and (slashidx == -1 or slashidx > colonidx):
+            parse_result = urlparse('ssh://' + git_url.replace(':', '/', 1))
+        else:
+            parse_result = urlparse(git_url)
+
         if parse_result.path and '@' in parse_result.path:
             git_url, version = git_url.rsplit('@', 1)
 


awelzel:
Hey Christian,

...the description here mentions:

> Python's urlparse parses SSH git URLs ("git@github.com/foo/bar") crudely, 

and #107 has the following example:

> When using a package source URL that includes a user specification (e.g. "git@github.com:user/repo.git") [...]

Not the `:` difference - the latter is what Github shows in the clone dialog, the former seem to be treated as a local URL.

Looking around a bit, this seems relevant from the docs of `git-clone`:
```
GIT URLS
[...]
       The following syntaxes may be used with them:
       •   ssh://[user@]host.xz[:port]/path/to/repo.git/
       [...]

       An alternative scp-like syntax may also be used with the ssh protocol:
       •   [user@]host.xz:path/to/repo.git/

       This syntax is only recognized if there are no slashes before the first colon. This helps differentiate a
       local path that contains a colon. For example the local path foo:bar could be specified as an absolute path or
       ./foo:bar to avoid being misinterpreted as an ssh url.
```

I was staring long enough, so tried a different fix. How about something as follows that recognizes the scp-style syntax?
```
--- a/zeekpkg/manager.py
+++ b/zeekpkg/manager.py
@@ -406,7 +406,19 @@ class Manager(object):
 
         # Support @ in the path to denote the "version" to checkout
         version = None
-        parse_result = urlparse(git_url)
+
+        #
+        # Prepend 'ssh://' and replace the first ':' with '/' if git_url
+        # looks like a scp-like URL, e.g. git@github.com:user/repo.git.
+        # urlparse will otherwise parse everything into path and the @
+        # is confusing the versioning logic.
+        #
+        colonidx, slashidx = git_url.find(':'), git_url.find('/')
+
+        if '://' not in git_url and colonidx > 0 and (slashidx == -1 or slashidx > colonidx):
+            parse_result = urlparse('ssh://' + git_url.replace(':', '/', 1))
+        else:
+            parse_result = urlparse(git_url)

```

Minmal smoke testing indicates that helps:

```
$ PYTHONPATH=. ./zkg  --extra-source other=git@github.com:ckreibich/packages.git search af_packet
other/j-gras/bro-af_packet-plugin - This plugin provides native AF_Packet support for Zeek.
other/j-gras/zeek-af_packet-plugin - This plugin provides native AF_Packet support for Zeek.
zeek/j-gras/bro-af_packet-plugin - This plugin provides native AF_Packet support for Zeek.
zeek/j-gras/zeek-af_packet-plugin - This plugin provides native AF_Packet support for Zeek.
```

ckreibich:
Thanks for looking, Arne. Yours is more precise, let's use it. I'd suggest that since the conditions are a bit tricky we call out "This syntax is only recognized if there are no slashes before the first colon" in the comment, since it explains the `slashidx > colonidx` condition.

I was torn about `git_url.replace(':')` since in case of `github.com:/foo/bar` (which works and is handy — you can say something like `$GIT_REPO/foo/bar` and have the variable be of any form) it replaces to `//foo/bar`, but it doesn't matter here. I noticed that the `ssh://` presence alone nudges urlparse to identify a netloc, but a slightly wrong one, so I think that's inferior:
```
12:44:09 10> urlparse('ssh://git@github.com:foo/bar')
12:44:14 10< ParseResult(scheme='ssh', netloc='git@github.com:foo', path='/bar', params='', query='', fragment='')
```
The only corner case I can still identify is a filesystem path whose first directory contains a colon (i.e. `git clone foo:bar/baz`). I think we can live with that!

I'll push an update in a second, feel free to review.

awelzel:
Thanks, LGTM.

PR #111, opened 2021-06-24
Fix crash in case the user enters just "zkg template"

ckreibich:
@bbannier flagged a backtrace when the user didn't go straight to a `zkg template` subcommand ... which apparently I never noticed. :facepalm: 

The output of "zkg template" (which will gain various commands over time) now mirrors that when just saying "zkg". Also, the help output in `zkg template --help` no longer double-mentions "template".

### DIFF
--- a/zkg
+++ b/zkg
@@ -2646,8 +2646,9 @@ def argparser():
         'template', help='Manage package templates.')
 
     template_command_parser = sub_parser.add_subparsers(
-        title='template commands',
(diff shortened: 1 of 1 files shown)


bbannier (approved):
Does it make sense to add a test baseline for the output of `zkg template` and `zkg template -h`?

ckreibich:
Probably. :-) It'd be nice not to have to spell out specific invocations, since technically `argparse` gives you enough structure to test invocations. Thanks for the approval.

PR #113, opened 2021-06-29
Added resilience when specifying local paths as package git repos

### LINKED ISSUE #112: Uncaught exception when installing local path that's not a git repo
This isn't great — would be better to just tell the user that they have to be in a git repo toplevel for this to work.
```
$ mkdir foo
$ cd foo
$ zkg install .
Traceback (most recent call last):
  File "/home/christian/inst/opt/zeek/bin/zkg", line 2704, in <module>
    main()
  File "/home/christian/inst/opt/zeek/bin/zkg", line 2701, in main
    args.run_cmd(manager, args, config, configfile)
  File "/home/christian/inst/opt/zeek/bin/zkg", line 517, in cmd_install
    if is_git_repo_dirty(name):
  File "/home/christian/inst/opt/zeek/bin/zkg", line 331, in is_git_repo_dirty
    repo = git.Repo(git_url)
  File "/usr/lib/python3.9/site-packages/git/repo/base.py", line 181, in __init__
    raise InvalidGitRepositoryError(epath)
git.exc.InvalidGitRepositoryError: /home/christian/t7/foo
```

ckreibich:
In addition to failing when a given path doesn't exist, the git.Repo() constructor also fails when a given path exists but is not a git
repository. That scenario caused backtraces with package installs, testing, and bundling when providing such invalid local paths.

We now check for this case and provide according output. This adds guards for the install, test, and bundling commands, which are the entry points for providing local URLs.

This also expands the test case for invalid repos.

Resolves #112.


zkg
```diff
@@ -453,9 +469,13 @@ def cmd_test(manager, args, config, configfile):
...
-        if is_git_repo_dirty(name):
-            print_error('error: local git clone at {} is dirty'.format(name))
-            sys.exit(1)
+        if is_local_git_repo_url(name):
+            if not is_local_git_repo(name):
+                print_error('error: path {} is not a git repository'.format(name))
+                sys.exit(1)
+            if is_local_git_repo_dirty(name):
+                print_error('error: local git clone at {} is dirty'.format(name))
+                sys.exit(1)
```
rsmmr, line 478: How about one more wrapper function for this block, which is repeated 3 times across the commands.

ckreibich: Yeah good point, thanks Robin. Done. We may want to start a `zeekpkg/_gitutil.py` soon, just for all the git-related utility functions...

rsmmr (approved):

PR #115, opened 2021-07-02
Require zeek-config for two tests that will fail without it

ckreibich:
"zkg test" constructs a suitable test environment via zeek-config's output.

It also needs zeek-config when constructing package dependencies, so it can verify whether packages depending on specific Zeek versions will work. So I'm surprised the absence doesn't trip up more tests. :grimacing:

### DIFF
--- a/testing/tests/bindir-install
+++ b/testing/tests/bindir-install
@@ -1,3 +1,6 @@
+# "zkg test" internally requires zeek-config
+# @TEST-REQUIRES: type zeek-config
+
 # @TEST-EXEC: bash %INPUT
 #
 # @TEST-EXEC: zkg test foo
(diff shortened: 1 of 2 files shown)


rsmmr (approved):
Looks good. 

I'm wondering if `zkg test` should abort if it can't locate `zeek-config` (I'm guessing it doesn't do that currently as otherwise these tests would have failed already.)

ckreibich:
Yeah, almost, Robin — `zkg test` _does_ abort when it cannot find `zeek-config`: https://github.com/zeek/package-manager/blob/master/zeekpkg/manager.py#L2289-L2305

Accordingly some of the btests do fail when `zeek-config` isn't available. My take on it was that this shouldn't be a test failure becuse there's nothing wrong with `zkg` — it's just that the tests require it to be there. We could add a test that verifies that `zkg test` indeed fails when it is unavailable.

I'll go ahead and merge, since we can easily revisit this.

PR #116, opened 2021-08-05
Accept `uninstall` as an alias for `remove`.

bbannier:


bbannier:
This might be a controversial patch, but I keep running into the package removal command being `remove`, not `uninstall`.

ckreibich (approved):
Not controversial from my side :-), especially given that we already have `bundle`/`unbundle`, `pin`/`unpin`, `load`/`unload`. Plus I can never remember the right command name for the various package managers either, so it makes sense to me to help the user.

ckreibich:
There's a small hiccup here because the docs build fails — apparently a [known issue](https://github.com/alex-rudakov/sphinx-argparse/issues/108) in `sphinx-argparse`. I'll look into resolving that one first before merging this one.

bbannier:
> There's a small hiccup here because the docs build fails — apparently a [known issue](https://github.com/alex-rudakov/sphinx-argparse/issues/108) in `sphinx-argparse`. I'll look into resolving that one first before merging this one.

I never even tried doing this. It would be great if docs build failures could be caught in a test somewhere, latest in a PR check to reduce the risk of merging broken stuff by accident.

ckreibich:
Yeah we could add this to CI. I only noticed because the docs update runs as part of `update-changes`, which I applied after I'd (locally) merged your branch.

PR #126, opened 2021-11-04
Request at least Sphinx 2.0 to avoid a dependency problem in RTD

ckreibich:
This resolves the build problems due to sphinx-doc/sphinx#9727 we've been seeing e.g. here:
https://readthedocs.org/projects/bro-package-manager/builds/15177099/

Working build here: https://readthedocs.org/projects/bro-package-manager/builds/15177411/


rsmmr (approved):
@bbannier I think we need this for Spicy as well.

PR #133, opened 2022-04-26
Make Zeek-provided CMake modules available to packages.

bbannier:
With this patch we make Zeek-provided CMake modules discoverable by
CMake invocations in package script instructions, specifically we append
the directory exposed by `zeek-config --cmake_dir` in the environment
variable `CMAKE_PREFIX_PATH`. If the environment variable is already set
we append to it, so any existing settings take precedence. If script
instructions modify this variable they will find the same CMake modules
if they prepend to the environment variable.

We inject the environment variable by setting it for the whole `zkg`
process at a pretty high level so that it is inherited by any forked
processes like e.g., shells spawned for package script instructions.
Ideally we would use a more general mechanism to set up the environment
for script instructions so that we can use an identical mechanism for
e.g., `PATH`, `CMAKE_PREFIX_PATH` and any future additions.

We also add persistence for this setting by adding it to the zkg
`config` file under `paths.cmake_dir`. Before this patch all variables
in the `paths` section referred to target paths for user settings or
packages (state, script, plugin directory), so this might not be the
optimal ultimate format to store this information.

### DIFF
--- a/zeekpkg/manager.py
+++ b/zeekpkg/manager.py
@@ -231,7 +231,7 @@ class Manager(object):
             load all installed packages that have been marked as loaded.
     """
 
-    def __init__(self, state_dir, script_dir, plugin_dir, zeek_dist='',
+    def __init__(self, state_dir, script_dir, plugin_dir, cmake_dir, zeek_dist='',
                  user_vars=None, bin_dir=''):
         """Creates a package manager instance.
 
@@ -241,6 +241,8 @@ class Manager(object):
             script_dir (str): value to set the `script_dir` attribute to
 
             plugin_dir (str): value to set the `plugin_dir` attribute to
+
+            cmake_dir (str): value to set the `cmake_dir` attribute to
 
             zeek_dist (str): value to set the `zeek_dist` attribute to
 
@@ -261,6 +263,7 @@ class Manager(object):
         # The bro_dist attribute exists just for backward compatibility
         self.bro_dist = zeek_dist
         self.zeek_dist = zeek_dist
+        self.cmake_dir = cmake_dir
         self.state_dir = state_dir
         self.user_vars = {} if user_vars is None else user_vars
         self.backup_dir = os.path.join(self.state_dir, 'backups')
--- a/zkg
+++ b/zkg
@@ -277,11 +277,22 @@ def create_config(args, configfile):
     if not zeek_dist:
         zeek_dist = get_option(config, 'paths', 'bro_dist', '')
 
+    cmake_dir = get_option(config, 'paths', 'cmake_dir', '')
+
+    if cmake_dir:
+        # If a cmake_dir is set, inject it into any CMake invocations created
+        # from this processess via `CMAKE_PREFIX_PATH`. This allows packages to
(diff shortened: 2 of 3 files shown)


bbannier:
@ckreibich, this is the WIP implementation we talked about offline. This seems to get the job done for me, but is more an exploration on whether addressing this in zkg itself would be worthwhile (as opposed to e.g., script instructions consuming output from `zeek-config` directly.

ckreibich (approved):

ckreibich:
I think this looks fine, Benjamin, especially since there's already that `zeek-config` invocation in there to collect other paths. (If you can think of a test case that actually checks the `CMAKE_PREFIX_PATH` that would be ideal.)

I'm not sure what else we might do here — I guess for the Zeek-bundled `zkg` we could populate the appropriate paths for a given Zeek installation, but then we'd still need this to work when the user installs `zkg` independently, and for that `zeek-config` strikes me as the right way to do it.

zeekpkg/manager.py
```diff
@@ -231,7 +231,7 @@ class Manager(object):
             load all installed packages that have been marked as loaded.
     """
 
-    def __init__(self, state_dir, script_dir, plugin_dir, zeek_dist='',
+    def __init__(self, state_dir, script_dir, plugin_dir, cmake_dir, zeek_dist='',
```
awelzel, line 234: We're using `zeekpkg.Manager` and this is an API change in that `cmake_dir` is now required to construct a `Manager`. I haven't looked at the rest, but can it be defaulted to an empty string? In the environment I'm having in mind, there's also no reasonable `cmake_dir` available to be provided.

Would also suggest to add it after the last parameter, otherwise the API changes when using positional arguments only. Though I really hope keyword args are used for construction elsewhere.

bbannier: I think this patch is not the correct fix, and we will not add it in this form.

zkg
```diff
@@ -277,11 +277,22 @@ def create_config(args, configfile):
...
 
+    cmake_dir = get_option(config, 'paths', 'cmake_dir', '')
+
+    if cmake_dir:
+        # If a cmake_dir is set, inject it into any CMake invocations created
+        # from this processess via `CMAKE_PREFIX_PATH`. This allows packages to
+        # use Zeek-provided CMake modules. We append to any existing settings
+        # in the caller's environment, so existing setting keep precedence.
+        cmake_prefix_path = os.environ.get('CMAKE_PREFIX_PATH', '')
+        os.environ['CMAKE_PREFIX_PATH'] = ';'.join([cmake_prefix_path, cmake_dir])
```
awelzel, line 288: Suggesting to use `os.pathsep.join([...])` rather than `;`

bbannier: This is not a path, but a list of paths. CMake encodes lists as strings with entries separated by `;`.

zeekpkg/manager.py
awelzel: Alright, thanks :-)

zkg
awelzel: Oh. I didn't know. that. Thanks for the pointer.

PR #151, opened 2023-01-09
Add pre-commit config

bbannier:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - id: trailing-whitespace
+    exclude: testing/baselines
+  - id: end-of-file-fixer
+    exclude: testing/baselines
+  - id: check-yaml
+  - id: check-added-large-files
+- repo: https://github.com/psf/black
+  rev: 23.1a1
+  hooks:
+  - id: black
```
ckreibich, line 16: The black docs say that it is recommended to set a language version here: https://black.readthedocs.io/en/stable/integrations/source_version_control.html  ... it's not clear to me how that matters since Black shouldn't be that selective, and it introduces one more spot where the language version is pinned down. Do you happen to know if it matters?

ckreibich (changes requested):
I look forward to having this in there. One request Benjamin: could you introduce a `.git-blame-ignore-revs` file covering the reformatting commit, so we keep `git blame` working? We have that in Zeek too.

https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

setup.py
```diff
@@ -1,28 +1,28 @@
...
-    long_description=open('README').read(),
-    license='University of Illinois/NCSA Open Source License',
-    keywords='zeek zeekctl zeekcontrol package manager scripts plugins security',
-    maintainer='The Zeek Project',
-    maintainer_email='info@zeek.org',
-    url='https://github.com/zeek/package-manager',
-    scripts=['zkg'],
-    packages=['zeekpkg'],
+    name="zkg",
+    version=open("VERSION").read().replace("-", ".dev", 1).strip(),
```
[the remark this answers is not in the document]
ckreibich: Okay so given this it looks like CodeQL is going to complain about stuff every time it comes up again in a new place as part of a code change, even if the alert has previously been whitelisted, like this one. That's exactly why the idea of keeping the whitelisting out of the code is not helpful. I was curious whether it'd be smarter than that ... and am now thinking we should remove it altogether.

.pre-commit-config.yaml
bbannier: I had to look up the reasoning for that, https://github.com/psf/black/pull/2430. In a quick test this seems to lead to `black` always requiring that the an installation for the given version is available on the host running `pre-commit`. I don't think we want to require that users install python3.7 out of band to be able to use the hooks.

setup.py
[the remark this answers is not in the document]
bbannier: The proper fix for the warnings on `open` is probably to implement one of the suggested workarounds. I added a patch doing that.

doc/conf.py
```diff
@@ -119,23 +119,24 @@
...
 
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
+on_rtd = os.environ.get("READTHEDOCS", None) == "True"
 
 if not on_rtd:
     # only import and set the theme if we're building docs locally
     import sphinx_rtd_theme
-    html_theme = 'sphinx_rtd_theme'
+
+    html_theme = "sphinx_rtd_theme"
```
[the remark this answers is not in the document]
bbannier: Since Sphinx seems to use magic variables, I am unsure what pattern to use to suppress this warning. Might be worthwhile just manually dismissing this whenever we make an edit to this file. That way we'd still get CodeQL analysis everywhere else (where we are unlikely to use such weird patterns).

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Migrate code style to Black
+327835fae527bc63159a8bdd904b1546594e0329
```
awelzel, line 2: Can you do this in the same commit as the reformatting? Isn't that predicting the future then? (Kudos if you can pull that off)

bbannier: Overeager rebase and squashing 🤦. I'll move this to its own commit to avoid having to compute the hash.

setup.py
```diff
@@ -2,11 +2,22 @@
...
+
+def long_description():
+    with open("README") as f:
+        return f.read()
+
+
 setup(
     name="zkg",
-    version=open("VERSION").read().replace("-", ".dev", 1).strip(),
+    version=version(),
```
awelzel, line 18: No need to change, just leaving it here as an alternative:

```
pathlib.Path("README").read_text().replace(...)
```

doc/ext/sphinxarg/ext.py
```diff
@@ -18,7 +18,7 @@
```
awelzel, line 18: Is the sphinxarg ext vendored? (https://sphinx-argparse.readthedocs.io/en/stable/install.html)

Suggest to exclude it from reformatting.

setup.py
bbannier: I like that more than the manual open calls. I'll still keep the separate functions though since they help reduce complexity in the args of `setup`.

zeekpkg/_util.py
```diff
@@ -278,14 +282,30 @@ def git_remote_urls(repo):
...
-        return False;
+        return False
 
     if len(s) != 40:
         return False
 
     for c in s:
-        if c not in {'a', 'b', 'c', 'd', 'e', 'f',
-                     '0', '1', '2', '3', '4', '5', '6', '7', '8', '9'}:
+        if c not in {
```
awelzel, line 291: Brrr... Maybe replace with a single return at the end of this function as follows or regex it?
```
hexdigits = set(string.hexdigits.lower())
return all([c in hexdigits for c in s])
```

bbannier: While I completely agree that this can be improved, this line appears in the diff due to automated edits. Let's keep non-automated changes out of this PR as much as possible. I'd rather drop this PR completely than address every issue in code touched by it.

awelzel: Sure, I do agree with that sentiment. This hunk was just standing out so prominently I left the comment. It's the only one I intended to point out. Please don't close the PR before merging it.

bbannier: Okay, pushed a patch fixing _just this one instance_ 😉

awelzel (approved):
Approving based on skimming through and I like the idea.

.pre-commit-config.yaml
ckreibich: Ah, definitely — thanks. Case closed.

zeekpkg/_util.py
ckreibich: Elegant fix!

ckreibich (approved):

PR #152, opened 2023-01-10
Remove CodeQL

ckreibich:
Let's use something with in-code annotation flexibility instead. I think none of CodeQL fits our workflow. Happy to discuss in this PR if anyone feels strongly about this.


awelzel:
> Let's use something with in-code annotation flexibility instead. I think none of CodeQL fits our workflow. Happy to discuss in this PR if anyone feels strongly about this.

Should we consider adding a minimal `flake8` pre-commit with this? Possibly after #151 then.

PR #153, opened 2023-01-17
Remove CodeQL, add flake8

awelzel:
@ckreibich , this builds on top of your #152 but at the same time adds a flake8 pre-commit hook.

Nothing critical stands out. A bit of logging broke during #151 , otherwise minor things around unused variables, import orderings and comparisons with bool.

Prefer to add some form of linter rather than having none at all.


bbannier (commented):
From skimming over the changes, this looks good to me.

IMO it would be a good idea to add additional linters to cover more scenarios here or as a follow-up, e.g., pylint and bandit.

awelzel:
> IMO it would be a good idea to add additional linters to cover more scenarios here or as a follow-up, e.g., pylint and bandit.

Yeah, not sure about about having flake8 and pylint due to possibly needing to appease them with different annotations, bandit could be interesting. There's also a bunch of flake8 plugins if we wanted to go wild: https://github.com/DmytroLitvinov/awesome-flake8-extensions

ckreibich (approved):
Thanks Arne, one less thing for me to do today.

PR #156, opened 2023-02-02
Fix zkg bundled usage

awelzel:
This reverts commit 2b11486b09d76df40aafd411aae88cb5cf902850 which naively moved imports above a sys.path.insert()... Adapt flake8 to ignore those import warnings and add a test so hopefully this won't happen again.


testing/tests/bundled-zkg
```diff
@@ -0,0 +1,21 @@
+# @TEST-DOC: Simulate bundling of zkg in Zeek and ensure zkg works. This got broken during a cleanup previously.
+# @TEST-EXEC: bash %INPUT > output
+# @TEST-EXEC: btest-diff output
+
+set -eu
+
+# Reset PATH and ensure we don't se the scripts/zkg wrapper.
+export PATH=$ORIGPATH
+if type zkg ; then
```
timwoj, line 9: I'm not sure this is correct. On a local docker instance using the same image we used on Cirrus, `type zkg` outputs `zkg is /opt/zeek/bin/zkg` and returns a `0` exit status. This causes the script to always go into the block and exit. Changing the condition here to be `! type zkg` makes it skip over the exit, but then the output file has the string output by `type` and not what's expected in the baseline.

timwoj (changes requested):

testing/tests/bundled-zkg
```diff
@@ -0,0 +1,21 @@
+# @TEST-DOC: Simulate bundling of zkg in Zeek and ensure zkg works. This got broken during a cleanup previously.
+# @TEST-EXEC: bash %INPUT > output
+# @TEST-EXEC: btest-diff output
+
+set -eu
+
+# Reset PATH and ensure we don't se the scripts/zkg wrapper.
```
timwoj, line 7:
```suggestion
# Reset PATH and ensure we don't see the scripts/zkg wrapper.
```

testing/tests/bundled-zkg
timwoj: It's really just failing because `/opt/zeek/bin` is in the PATH (and ORIGPATH, for that matter) and so it's finding that version of `zkg` over the one that it's trying to test.

awelzel: That `exit 1` was there to ensure there wouldn't be a `zkg` installed, but clearly the environment is different and in retrospect was more harmful.

I've pushed a fixup to print `command -v zkg` as a pre-check it's picking up the one inside install-dir.

:crossed_fingers:

ckreibich (approved):

PR #159, opened 2023-04-21
Change --force to fail on failing tests

### LINKED ISSUE #137: Consider always failing `install` if not skipped tests fail
One can currently use `install`'s `--force` flag to disable interactive confirmations,
```
--force              Don't prompt for confirmation or user variables.
```

One uses this flag to skip the interactive confirmation during normal install:
```console
$ zkg install spicy-plugin
  zeek/zeek/spicy-plugin (v1.3.17)

Proceed? [Y/n]
```

This has the possibly surprising effect of installing the package even if its `test_command` failed. I would argue this is surprising since one can already skip tests during installation with `--skiptests` if they are e.g., known to fail.

If `build_command` fails the installation process fails with an error code, and it might be much simpler if `test_command` behaved the same way. This would allow to use `--force` for non-interactive installations. Currently this requires scripting the possible interactive queries, e.g., to say `Y` to the confirm the installation and `N` to abort after failed tests one would use

```console
$ echo 'Y\nN' | zkg install spicy-plugin
```
Without `N` here `zkg` would forever wait for input while `--force` would imply always `Y`.
If a failing `test_command` would abort the installation one could simply use
```console
$ zkg install spicy-plugin --force
```

awelzel:
This is a user-visible change in behavior, but to something sane, so
hopefully that's acceptable.

Currently, `--force` is oftentimes used in scenarios where it should mean
"non-interactive". At the same time, `--force` also implies to ignore
failing test commands. This is not reasonable behavior, particularly as
`--skiptests` is available. This patch changes the behavior and causes
install/upgrade to exit with a failure code when tests fail even with
`--force` provided.

If someone intentionally wants to install a package with broken tests,
they can/should use `--skiptests` explicitly instead. That is less
surprising and stands out more.

Fixes #137


awelzel:
This also sneaks in being a bit more conservative about the specified resources.

zkg
```diff
@@ -726,7 +726,7 @@ def cmd_install(manager, args, config, configfile):
 
             if "test_command" not in info.metadata:
                 zeekpkg.LOG.info(
-                    'Skipping unit tests for "%s": ' "no test_command in metadata", name
+                    'Skipping unit tests for "%s": no test_command in metadata', name
```
bbannier, line 729: Could just go with an f-string here.

zkg
```diff
@@ -1469,7 +1469,16 @@ def cmd_upgrade(manager, args, config, configfile):
...
+            # Get info object for the next version as there may have been a
+            # test_command added during the upgrade.
+            next_info = manager.info(name, version=version, prefer_installed=False)
+            if next_info.invalid_reason:
+                print_error(
+                    f'error: invalid package "{name}": {next_info.invalid_reason}'
+                )
+                sys.exit(1)
+
+            if "test_command" not in next_info.metadata:
```
bbannier, line 1481: This is slightly hacky since one always needs to remember that the `info` from `outdated_packages` might be outdated (which lead exactly to the bug you are fixing here). I wonder whether putting e.g., `[installed_version, next_version]` into `outdated_packages` would be harder to misuse. That would also help keeping that extra error handling code you added here in a single place.

bbannier (approved):
This looks good to me from the functionality side, only real issue is on the fix for the upgrade path you sneaked in (maybe that should be its own commit anyway).

zkg
awelzel: I've split the fix into a separate commit, but otherwise skipped changing anything around here. That doesn't mean I'm overly happy with that is, but any improvements would involve changing `outdated_packages` to actually contain namedtuples and I didn't want to go there.

ckreibich (approved):
Yeah looks good to me, thanks. I think `--force` has always been a poor choice because it's fairly ambiguous, and to me suggests something closer to skipping (or ignoring) tests, not anything prompt-related. I suggest we deprecate it and introduce something better — I was tempted to suggest `-y`/`--yes`/`--assume-yes`, but since we also use it to suppress user variable input, perhaps `--noprompts`. (Can be a subsequent PR.)

PR #161, opened 2023-05-25
Add --ignore-dirty-git option to test and install commands

### LINKED ISSUE #158: add flag to ignore dirty local git clone
When doing `zkg install`, a [check is made](https://github.com/zeek/package-manager/blob/5ade27d716fd65342be592abf277eea114fefaa7/zkg#L382-L384) to ensure the local repository isn't dirty. If it is, the install fails.

For development purposes, it would be convenient to have a flag named something like `--ignore-dirty-git` that would allow the install operation to proceed anyway.

e.g.

```
$ touch asdfasdf

$ zkg install .
error: local git clone at . is dirty

$ zkg install --ignore-dirty-git .
...
```

related to #145

timwoj:
This one has annoyed me in the past as well. This is still missing a test to cover it, but I'm working on that.

Fixes #158

### DIFF
--- a/zkg
+++ b/zkg
@@ -374,12 +374,12 @@ def is_local_git_repo_dirty(git_url):
     return repo.is_dirty(untracked_files=True)
 
 
-def check_local_git_repo(git_url):
+def check_local_git_repo(git_url, allow_dirty):
     if is_local_git_repo_url(git_url):
         if not is_local_git_repo(git_url):
             print_error(f"error: path {git_url} is not a git repository")
             return False
-        if is_local_git_repo_dirty(git_url):
+        if not allow_dirty and is_local_git_repo_dirty(git_url):
             print_error(f"error: local git clone at {git_url} is dirty")
             return False
 
@@ -549,7 +549,7 @@ def cmd_test(manager, args, config, configfile):
     package_infos = []
 
     for name in args.package:
-        if not check_local_git_repo(name):
+        if not check_local_git_repo(name, args.ignore_dirty_git):
             sys.exit(1)
 
         version = args.version if args.version else active_git_branch(name)
@@ -619,7 +619,7 @@ def cmd_install(manager, args, config, configfile):
     package_infos = []
 
     for name in args.package:
-        if not check_local_git_repo(name):
+        if not check_local_git_repo(name, args.ignore_dirty_git):
             sys.exit(1)
 
         version = args.version if args.version else active_git_branch(name)
@@ -889,7 +889,7 @@ def cmd_bundle(manager, args, config, configfile):
         new_pkgs = []
 
         for name, version in packages:
-            if not check_local_git_repo(name):
+            if not check_local_git_repo(name, args.ignore_dirty_git):
                 sys.exit(1)
 
             if not version:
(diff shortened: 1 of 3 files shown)


zkg
```diff
@@ -2512,6 +2518,12 @@ def argparser():
         " the latest version tag, or if a package has none,"
         ' the default branch, like "main" or "master".',
     )
+    sub_parser.add_argument(
```
bbannier, line 2521: We need a similar option for `test`.

It is pretty unfortunate that we need to add these flags for each and every command. Another option would be to add it once to `top_parser` so we'd have `zkg --ignore-git-deps install .` instead of `zkg install --ignore-git-deps .`, similarly to what is done right now for `--verbose`.

zkg
```diff
@@ -2512,6 +2518,12 @@ def argparser():
...
+        "--ignore-dirty-git",
+        action="store_true",
+        help="Allows installation of packages from 'dirty' git clones instead"
+        " of failing.",
```
bbannier, line 2525: Nit: A way to avoid this weird line break would be to use

```suggestion
        help=(
            "Allows installation of packages from 'dirty' git clones instead of failing."
        ),
```

Here and below.

bbannier (changes requested):

zkg
timwoj: > We need a similar option for test.

It's there. I added one for test, install, and bundle.

> It is pretty unfortunate that we need to add these flags for each and every command.

I agree, but it also felt weird adding it at the top level, but only having it be valid for 3 out of the 21 possible commands.

testing/tests/install-ignore-dirty
```diff
@@ -0,0 +1,16 @@
+# Test invalid package names and paths
+# @TEST-EXEC: bash %INPUT 2>output
+# @TEST-EXEC: btest-diff output
```
bbannier, line 3: Just making sure, do we need to check that the output is empty on top of being able to run from a dirty directory? To me it seems `output` is secondary and we could just drop it here.

testing/tests/install-ignore-dirty
```diff
@@ -0,0 +1,16 @@
...
+mkdir ./packages/dirtyrepo
+( cd ./packages/dirtyrepo && \
+  git init . && \
+  touch README && \
+  cp ${PACKAGES}/foo/zkg.meta zkg.meta && \
+  git add README zkg.meta && \
+  git commit -m "Initial commit")
+echo README > ./packages/dirtyrepo/README
+
+zkg --config=$CONFIG install --force --ignore-dirty-git ./packages/dirtyrepo
```
bbannier, line 16: In addition to `install`, let's add a similar btest for `test` and `bundle`.

bbannier (approved):

testing/tests/install-ignore-dirty
timwoj: Renamed the test to `ignore-dirty` and added tests for `test` and `bundle`.

testing/tests/install-ignore-dirty
timwoj: I removed the output but added a couple of `test`s to make sure the output files from install and bundle exist correctly.

PR #166, opened 2023-07-03
CI: switch to Github and our Docker images, automatically push to PyPI when pushing a git tag

ckreibich:
`zkg`'s CI was using Docker to install our binary images, which stems from a time when we didn't have our own Docker images and seems pretty roundabout these days. This switches the testing to using our Docker images, using a matrix of lts, latest, and the nightly, with the latter allowed to fail.

It also adds auto-pushing to PyPI on git tag push, mirroring zeek/btest#96. Successful test run in my fork to test.pypi [here](https://github.com/ckreibich/package-manager/actions/runs/5441011271/jobs/9894517158).


.cirrus.yml
```diff
@@ -1,41 +0,0 @@
...
-      path: "testing/**/tmp.tar.gz"
-  always:
-    upload_btest_xml_results_artifacts:
-      path: "testing/**/btest-results.xml"
-      type: text/xml
-      format: junit
-
-zeek_lts_task:
-  container:
-    dockerfile: ci/zeek/Dockerfile
```
awelzel, line 29: Could the whole `ci/` directory now go, too?

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    steps:
+      - name: Install build environment
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends cmake g++ libssl-dev libpcap-dev make
+      - uses: actions/checkout@v3
+      - name: Run unit tests
+        run: |
+          make test
```
awelzel, line 42: Only thought is that this hides the failure details within the artifacts. Elsewhere had promoted it's nicer if `btest -d` is directly included in the CI output.

Maybe adding `-d` in `testing/Makefile` is sufficient?

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+        include:
+          - repo: zeek
+            version: latest
+            experimental: false
+          - repo: zeek
+            version: lts
+            experimental: false
+          - repo: zeek-dev
+            version: latest
+            experimental: true
```
awelzel, line 27: Slight preference to using `continue_on_error` here so it's immediately clear what that bool stands for.

awelzel (approved):
Thanks, nice. Just nits.

bbannier:
Without having looked into the details, do we want to add code which cleans out the existing zkg installation in the image? I am mostly worried about CI not testing what users would be running due to something accidentally picking up binaries or Python modules from the installation.

Currently e.g., `testing/btest.cfg` doesn't control `PYTHONPATH` so we rely on them magic in zkg to pick the right modules to import. I don't think the fix should be to set `PYTHONPATH`, but instead make sure that exactly one version of the package-manager code is in the CI system.

ckreibich:
I think all of the stuff you guys flagged is now addressed. I'll merge this so it's closed out before I'm out. Can always iterate if you guys spot anything else.

PR #170, opened 2023-11-03
Fix Github edit links in top-right of HTML pages

awelzel:
These links 404'd because they omitted the file suffix, and they didn't really link to anything editable, just the raw file.

---

In this case, "suffix" was used which is set by the readthedocs theme in older versions [1]. Seem good to replace with page_source_suffix.

Relates to zeek/zeek-docs#222

[1] https://github.com/readthedocs/sphinx_rtd_theme/pull/1104

### DIFF
--- a/doc/_templates/breadcrumbs.html
+++ b/doc/_templates/breadcrumbs.html
@@ -5,7 +5,7 @@
 {% if pagename != "search" %}
   {% if display_github %}
     {% if github_version == "master" %}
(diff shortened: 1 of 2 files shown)


bbannier (approved):
LGTM, this is consistent with the Zeek change.

Do you want to fix the build against zeek-dev in this PR or will you create a follow-up issue?

awelzel:
> LGTM, this is consistent with the Zeek change.
> Do you want to fix the build against zeek-dev in this PR or will you create a follow-up issue?

Thanks - I'll merge just the docs change with a broken pipeline. There's more to that failing test than just a simple baseline/exec update :cry:

PR #173, opened 2023-11-03
manager: Support __zeek_plugin__ magic file

### LINKED ISSUE #171: Cannot unload/load plugins with Zeek 6.1 - __zeek_plugin__ not recognized
Zeek 6.1 now uses `__zeek_plugin__` as plugin magic file. We decided to go without backwards compat for the old `__bro_plugin__` within in Zeek assuming users should just use the `zkg` version bundled with Zeek.

However, we never added the code for `zkg` to actually recognize and handle the new `__zeek_plugin__` file. There was a commit, but it never made it in 76c0d456abd5321d99abfe036afbcb7666d8b31e (and we should continue to handle `__bro_plugin__` in a backwards compatible manner.

Reference:
https://github.com/zeek/zeek/pull/3155#discussion_r1261222306

> The two scenarios that comes up:
>
> [...]
> * a newer zkg not recognizing __bro_plugin__ anymore. This seems like we we can/should add backwards compat on the zkg side for that, so that it moves either __bro_plugin__ or __zeek_plugin__ depending on which exists.

Effectively, with Zeek 6.1 currently plugins cannot be unloaded/loaded. There are actually tests failing in #170, but we never realized as it's not running regularly. Probably need an integration test on the Zeek side for zkg, similarly how we do things for zeekctl.

---
In a Zeek 6.1 container after installing zeek-more-hashes
```
# zkg install https://github.com/zeek/zeek-more-hashes
  https://github.com/zeek/zeek-more-hashes (v0.2.0)

Proceed? [Y/n] Y
Running unit tests for "https://github.com/zeek/zeek-more-hashes"
Installing "https://github.com/zeek/zeek-more-hashes".............
Installed "https://github.com/zeek/zeek-more-hashes" (v0.2.0)
Loaded "https://github.com/zeek/zeek-more-hashes"
# zeek -N | grep More
Zeek::MoreHashes - More analyzers and built-in functions for hashing. (dynamic, version 0.2.0)
# zkg unload zeek-more-hashes
The following packages will be UNLOADED:
  zeek/zeek/zeek-more-hashes

Proceed? [Y/n] Y
Unloaded "zeek/zeek/zeek-more-hashes"
# zeek -N | grep More
Zeek::MoreHashes - More analyzers and built-in functions for hashing. (dynamic, version 0.2.0)
```

awelzel:
Zeek 6.1 switched from __bro_plugin__ to __zeek_plugin__. Modify _write_plugin_magic() to handle the new and the old file.

Closes #171

### DIFF
--- a/testing/tests/plugin
+++ b/testing/tests/plugin
@@ -3,21 +3,21 @@
 # @TEST-EXEC: bash %INPUT
 # @TEST-EXEC: zkg install rot13
 
-# @TEST-EXEC: test -f plugins/packages/rot13/__bro_plugin__
+# @TEST-EXEC: test -f plugins/packages/rot13/__bro_plugin__ || test -f plugins/packages/rot13/__zeek_plugin__
 # @TEST-EXEC: btest-diff scripts/packages/rot13/__load__.zeek
 
 # Unloading the package should also disable the plugin, which we
 # detect via the renamed __bro_plugin__ magic file.
 # @TEST-EXEC: zkg unload rot13
 
-# @TEST-EXEC: test ! -f plugins/packages/rot13/__bro_plugin__
-# @TEST-EXEC: test -f plugins/packages/rot13/__bro_plugin__.disabled
+# @TEST-EXEC: test ! -f plugins/packages/rot13/__bro_plugin__ && test ! -f plugins/packages/rot13/__zeek_plugin__
+# @TEST-EXEC: test -f plugins/packages/rot13/__bro_plugin__.disabled || test -f plugins/packages/rot13/__zeek_plugin__.disabled
 
 # (Re-)loading the package should also (re-)enable the plugin.
 # @TEST-EXEC: zkg load rot13
 
-# @TEST-EXEC: test -f plugins/packages/rot13/__bro_plugin__
-# @TEST-EXEC: test ! -f plugins/packages/rot13/__bro_plugin__.disabled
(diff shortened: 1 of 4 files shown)


testing/tests/plugin
```diff
@@ -3,21 +3,23 @@
...
+# @TEST-EXEC: test -f plugins/packages/rot13/__bro_plugin__ || test -f plugins/packages/rot13/__zeek_plugin__
 # @TEST-EXEC: btest-diff scripts/packages/rot13/__load__.zeek
 
 # Unloading the package should also disable the plugin, which we
 # detect via the renamed __bro_plugin__ magic file.
 # @TEST-EXEC: zkg unload rot13
 
 # @TEST-EXEC: test ! -f plugins/packages/rot13/__bro_plugin__
-# @TEST-EXEC: test -f plugins/packages/rot13/__bro_plugin__.disabled
+# @TEST-EXEC: test ! -f plugins/packages/rot13/__zeek_plugin__
```
bbannier, line 14: Nit: I'd go with the following for symmetry with all the other fallbacks (after all this is just a single check against multiple versions).

```zeek
# @TEST-EXEC: test ! -f plugins/packages/rot13/__bro_plugin__ && test ! -f plugins/packages/rot13/__zeek_plugin__
```

bbannier (approved):
:shipit:

testing/tests/plugin
awelzel: Thanks!

PR #174, opened 2023-11-03
Protect from some aliases misuses/gotchas

awelzel:


awelzel:
@bbannier - is this roughly what you think makes sense to avoid alias misuse?

zeekpkg/package.py
```diff
@@ -57,6 +57,15 @@ def is_valid_name(name):
     return True
 
 
+def validate_aliases(metadata_dict):
```
bbannier, line 60: Do we need a new validator here, or could we instead require that `alias` is a valid name and extend the `is_valid_name` validator? It seems that an alias would need to be a valid package name.

zeekpkg/package.py
```diff
@@ -57,6 +57,15 @@ def is_valid_name(name):
...
+    """Return empty string if package contains valid aliases only."""
+    for a in aliases(metadata_dict):
+        if not re.match(r"^[a-zA-Z0-9][+.-_a-zA-Z0-9]*[a-zA-Z0-9]$", a):
```
bbannier, line 63: Could you add comments here why we introduce certain requirements? E.g., why do we require aliases to not start with a digit, but allow `+`/`.`/`-` in the name which would be valid file names for `@load`, but invalid `module` identifiers?

zeekpkg/manager.py
```diff
@@ -2969,6 +2970,54 @@ def install(self, pkg_path, version=""):
...
+            if ipkg.package == pkg:
+                continue
+
+            qn = ipkg.package.qualified_name()
+
+            package_names[ipkg.package.name] = qn
+            for ipkg_alias in ipkg.package.aliases():
+                alias_names[ipkg_alias] = qn
+
+        # Does the new package's name is the same as an existing alias?
```
bbannier, line 3000:
```suggestion
        # Is the new package's name the same as an existing alias?
```

bbannier (commented):
Thanks, this looks like a good approach to avoiding surprises once we have all needed information.

zeekpkg/manager.py
```diff
@@ -2969,6 +2970,54 @@ def install(self, pkg_path, version=""):
...
+        if pkg.name in alias_names:
+            qn = alias_names[pkg.name]
```
bbannier, line 3002: Nit: we could reduce the number of lookups here (this would be even more compact with python-3.8 assignment expressions, but we need >=python-3.7).

```suggestion
        qn = alias_names.get(pkg.name)
        if qn:
```

Same for the lookups in the loop below.

awelzel: If we had `:=` available, yes. Otherwise I find the `.get()` not as pretty.

zeekpkg/package.py
awelzel: Hmm.. This was mostly thinking "reasonably sane package", but I also realized we do not have that elsewhere.

Let me revert to simply checking for `/` and fold that into `is_valid_name()`.

bbannier (approved):
Thanks! 🚢

PR #175, opened 2024-01-23
Add developer docs about zkg's internal use of directories

ckreibich:
I'm putting together some developer docs since I'm hoping to find more time to get back to working on `zkg`.

### DIFF
--- a/doc/developers.rst
+++ b/doc/developers.rst
@@ -88,3 +88,116 @@ INI File Option            ```name```                      `name`
 ========================== =============================== ===========================
 
 Python API docstrings roughly follow the `Google Style Docstrings`_ format.
+
+Internals
+---------
+
+``zkg``'s view of a package
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``zkg`` maintains copies of a Zeek package in up to four places:
+
+- A long-lived clone of the package's git repo in
+  ``$STATEDIR/clones/package/<name>``. This clone (not its installed version,
+  see below) is ``zkg``'s "authoritative" view of a package.
+
+- A conceptually short-lived clone in ``$STATEDIR/scratch/<name>``, for
+  retrieving information about a package ("short-lived", because access to those
+  copies is momentary -- not because ``zkg`` necessarily cleans up after those
+  copies).
+
+- A "stage". A stage is any place in which ``zkg`` actually installs a package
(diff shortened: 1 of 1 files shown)


doc/developers.rst
```diff
@@ -88,3 +88,116 @@ INI File Option            ```name```                      `name`
...
+When installing a package's ``script_dir`` or ``plugin_dir`` into a staging area
+and the source file is a tarfile, ``zkg`` temporarily extracts the tarball into
+``$STATEDIR/scratch/untar/``.
+
+There's little or no cleanup of files in the scratch space after the operations
+creating them complete. ``zkg`` only deletes, then (re-)creates, the directories
+involved upon next use.
+
+When ``zkg`` isn't in the middle of executing any commands, you can always
+delete the scratch space without negatively affecting ``zkg``.
```
awelzel, line 187: I guess that's on your mind, inverting this behavior and introducing `--keep-scratch` / `--keep-testing` or some `--keep-temporary` wildcard would be less surprising.

awelzel (approved):

doc/developers.rst
```diff
@@ -88,3 +88,116 @@ INI File Option            ```name```                      `name`
...
+
+- A conceptually short-lived clone in ``$STATEDIR/scratch/<name>``, for
+  retrieving information about a package ("short-lived", because access to those
+  copies is momentary -- not because ``zkg`` necessarily cleans up after those
+  copies).
+
+- A "stage". A stage is any place in which ``zkg`` actually installs a package
+  for running in Zeek. This can be the local Zeek installation, or locally below
+  ``$STATEDIR/testing`` when testing a package. So "staging" here does not mean
+  "not yet live"; rather the opposite: it's closer to "any place where the
```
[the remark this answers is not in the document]
ckreibich: That's just my read of it btw, but I don't see any code where a stage would be transitioned to a deployment, so I think that interpretation is closest. And yeah, I'd be happy to change the code. Simply calling it an `Installation` might do it.

doc/developers.rst
ckreibich: Yeah, especially for `scratch`. `zkg` could no doubt clean up properly in there. And for `testing` it could at least clean up when the tests pass. (And it should be much better about pointing you at the trouble if they don't, but that's another one.)

PR #183, opened 2024-03-01
Standardize on ruff for python formatting/linting via pre-commits

timwoj:
This is part of a minor effort to standardize all of our python pre-commits on the same tooling, in this case [ruff](https://github.com/astral-sh/ruff). It adds one minor formatting fix and fixes a few findings from the linter (some by just suppressing them).


bbannier (approved):
https://github.com/zeek/btest/pull/101#pullrequestreview-1931819444

PR #188, opened 2024-04-13
Modernize project

bbannier:
This PR contains a number of cleanups of the project.

- Move most project config from `setup.py` to `pyproject.toml`: with this one can use modern tools like e.g., rye or hatch to manage dev setups for this project
- Move formatting/linting to `ruff`/`ruff-format`: we have already done this for a number of other projects
- Activate additional lints. These seemed overall non-intrusive to me and with low rate of false positives (`ruff` could also fix most of them automatically). This list could be expanded, but I stopped here.
- Misc. cleanups


timwoj (approved):
I looked over all of it but most of it looks like automated changes. I didn't see anything out of place. The change to pyproject.toml looks good to me too.

PR #189, opened 2024-05-08
Make requirements.txt match pyproject file

timwoj:
Notably, the two versions listed here are the current releases. The current setting causes RTD to fail since it can't find versions newer than those.


bbannier (approved):
Thanks for catching this! 🚀

PR #204, opened 2025-04-25
Enable lints around return statements

bbannier:
This enables a class of return-related lints[^1] around return
statements. I mainly came here for redundant returns after `else`[^2],
but they all seemed useful. In general this seems to reduce nesting, and
also make overall flow clearer.

[^1]: https://docs.astral.sh/ruff/rules/#flake8-return-ret
[^2]: https://docs.astral.sh/ruff/rules/superfluous-else-return/

### DIFF
--- a/zeekpkg/manager.py
+++ b/zeekpkg/manager.py
@@ -1318,15 +1318,17 @@ class Manager:
         if ipkg.status.tracking_method == TRACKING_METHOD_VERSION:
             version_tags = git_version_tags(clone)
             return self._install(ipkg.package, version_tags[-1])
-        elif ipkg.status.tracking_method == TRACKING_METHOD_BRANCH:
+
+        if ipkg.status.tracking_method == TRACKING_METHOD_BRANCH:
             git_pull(clone)
             return self._install(ipkg.package, ipkg.status.current_version)
-        elif ipkg.status.tracking_method == TRACKING_METHOD_COMMIT:
+
+        if ipkg.status.tracking_method == TRACKING_METHOD_COMMIT:
             # The above check for whether the installed package is outdated
             # also should have already caught this situation.
             return "package is not outdated"
-        else:
-            raise NotImplementedError
+
+        raise NotImplementedError
 
     def remove(self, pkg_path):
         """Remove an installed package.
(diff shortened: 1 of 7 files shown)


zeekpkg/template.py
```diff
@@ -483,7 +483,7 @@ def contentdir(self):
         Returns:
             str: relative path to the content directory
         """
-        return None
```
evantypanski, line 486: These blank returns at the end of the methods look off, I'd prefer if they were just gone (though I know the changes were automated)

evantypanski (approved):

zeekpkg/template.py
bbannier: Good catch, thanks. I have replaced these with `pass`. I already made some manual edits here by adding empty lines between consecutive `if` blocks.

zeekpkg/manager.py
```diff
@@ -3201,12 +3204,14 @@ def _is_branch_outdated(clone, branch):
...
     if tracking_method == TRACKING_METHOD_VERSION:
         return _is_version_outdated(clone, ref_name)
-    elif tracking_method == TRACKING_METHOD_BRANCH:
+
+    if tracking_method == TRACKING_METHOD_BRANCH:
         return _is_branch_outdated(clone, ref_name)
-    elif tracking_method == TRACKING_METHOD_COMMIT:
+
+    if tracking_method == TRACKING_METHOD_COMMIT:
         return False
```
awelzel, line 3212: Might be very subjective, but the `if..elif..elif` seemed nicer . I guess this should be `match case` anyhow if we can, so just ignore this drive-by comment :-)

zeekpkg/template.py
```diff
@@ -483,7 +483,7 @@ def contentdir(self):
...
+        pass
```
awelzel, line 486:
```suggestion
        ...
```

Seems the current Python docs use ellipsis for the placeholder: https://docs.python.org/3/library/abc.html#abc.abstractmethod

bbannier: I'd prefer to not do that change now since nothing here says that this is an abstract method and ellipsis seem to be mostly used to denote a placeholder (e.g., to declare signatures `fn foo(): ...`). If we cleaned up `template.py` to declare proper abstract interfaces we should revisit that.

zeekpkg/manager.py
awelzel: Ah, Python 3.9. Sorry for the noise.. I'm resolving, do what you think makes sense :-)

zeekpkg/template.py
awelzel: Also resolving this one. Feel free to merge, seems you and Evan think this is worth it, so :+1:

awelzel (approved):

PR #209, opened 2025-08-11
Add type hints

bbannier:
This PR adds type hints to most functions, and adds a pre-commit hook running mypy to validate typing. I did not attempt to clean up any of the APIs, but having proper typing could be a first step for that. It also makes it easier to modify and extend code safely, and helped identify a few incorrect typings which would otherwise manifest as runtime errors.

### DIFF
--- a/zeekpkg/_util.py
+++ b/zeekpkg/_util.py
@@ -2,6 +2,7 @@
 These are meant to be private utility methods for internal use.
 """
 
+import configparser
 import errno
 import importlib.machinery
 import os
@@ -9,12 +10,14 @@ import shutil
 import string
 import tarfile
 import types
+from collections.abc import Callable, Iterable
+from typing import IO, TextIO, cast
 
 import git
 import semantic_version as semver
 
 
-def make_dir(path):
+def make_dir(path: str) -> None:
     """Create a directory or do nothing if it already exists.
 
     Raises:
@@ -30,7 +33,7 @@ def make_dir(path):
             raise
 
 
-def normalize_version_tag(tag):
+def normalize_version_tag(tag: str) -> str:
     """Given version string "vX.Y.Z", returns "X.Y.Z".
     Returns other input strings unchanged.
     """
@@ -40,7 +43,7 @@ def normalize_version_tag(tag):
     return tag
 
 
-def delete_path(path):
+def delete_path(path: str) -> None:
     if os.path.islink(path):
         os.remove(path)
         return
@@ -54,12 +57,16 @@ def delete_path(path):
         os.remove(path)
 
 
-def copy_over_path(src, dst, ignore=None):
+def copy_over_path(
+    src: str,
+    dst: str,
+    ignore: Callable[[str, list[str]], Iterable[str]] | None = None,
+) -> None:
     delete_path(dst)
     shutil.copytree(src, dst, symlinks=True, ignore=ignore)
 
 
-def make_symlink(target_path, link_path, force=True):
+def make_symlink(target_path: str, link_path: str, force: bool = True) -> None:
     try:
         os.symlink(target_path, link_path)
     except OSError as error:
@@ -70,7 +77,7 @@ def make_symlink(target_path, link_path, force=True):
             raise error
 
 
-def safe_tarfile_extractall(tfile, destdir):
+def safe_tarfile_extractall(tfile: str, destdir: str) -> None:
     """Wrapper to tarfile.extractall(), checking for path traversal.
 
     This adds the safeguards the Python docs for tarfile.extractall warn about:
@@ -88,7 +95,7 @@ def safe_tarfile_extractall(tfile, destdir):
         Exception: if the tarfile would extract outside destdir
     """
 
-    def is_within_directory(directory, target):
+    def is_within_directory(directory: str, target: str) -> bool:
         abs_directory = os.path.abspath(directory)
         abs_target = os.path.abspath(target)
         prefix = os.path.commonprefix([abs_directory, abs_target])
@@ -103,7 +110,7 @@ def safe_tarfile_extractall(tfile, destdir):
         tar.extractall(destdir)
 
 
-def find_sentence_end(s):
+def find_sentence_end(s: str) -> int:
     beg = 0
 
     while True:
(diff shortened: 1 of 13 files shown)


zeekpkg/manager.py
```diff
@@ -1126,10 +1126,8 @@ def _refresh_source(self, name, aggregate=False, push=False):
                     try:
                         clone = git_clone(url, clonepath, shallow=True)
                     except git.GitCommandError as error:
-                        LOG.warn(
-                            "failed to clone %s, skipping aggregation: %s",
-                            url,
-                            error,
+                        LOG.warning(
+                            f"failed to clone {url}, skipping aggregation: {error}",
```
awelzel: No strong opinion for zkg, but for logging calls, seems it's still recommended to avoid f-strings elsewhere (reddit has enough hits, eventually leading to this)

https://google.github.io/styleguide/pyguide.html#3101-logging

Not sure that matters for zkg, but in certain environments  querying on the uninterpolated string could get all occurrences of a warning, regardless of which url/error was triggered. That stops working with `f-strings` because the non-interpolated string isn't available anymore.

zeekpkg/manager.py
```diff
@@ -1378,7 +1369,7 @@ def remove(self, pkg_path):
                     LOG.debug("removing link %s", link)
                     os.unlink(link)
                 except OSError as err:
-                    LOG.warn("cannot remove link for %s", err)
+                    LOG.warning(f"cannot remove link for {err}")
```
awelzel: ...and the f-string pattern becomes inconsistent with `LOG.debug("removing link %s", link)` above (just happened to spot this).

zeekpkg/_util.py
```diff
@@ -206,19 +218,20 @@ def git_default_branch(repo):
...
+                remote_prefix = "origin/"
 
-            if head_ref_name.startswith(remote_prefix):
-                return head_ref_name[len(remote_prefix) :]
+                if head_ref_name.startswith(remote_prefix):
+                    return head_ref_name[len(remote_prefix) :]
 
-            return head_ref_name
+                return head_ref_name
+        except Exception:
```
awelzel, line 230: This now potentially catches `AssertExceptions` raised by the `asserts` above (and nothing else?) which feels pretty off to me for Python:
```
In [2]: try:
   ...:     assert 1 == 2
   ...: except Exception:
   ...:     print("caught")
   ...: 
caught
```

I'd just remove the assert tbh, the original code seemed fine.

awelzel: Hmm, are these asserts some sort of mypy hints? Then ignore my comment. This is just new to me.

Still think if there's a more direct way to indicate the expected type like 
`ref = typing.cast(git.Reference, repo.head.ref)` - seems to do the trick here - we should probably prefer that.

I'm just picking up on the potential AssertionError raised coupled with the unfortunate `except Exception` below.

zeekpkg/_util.py
```diff
@@ -206,19 +218,20 @@ def git_default_branch(repo):
...
-            head_ref_name = remote.refs.HEAD.ref.name
-        except Exception:
-            head_ref_name = None
-
-        if head_ref_name:
-            remote_prefix = "origin/"
+            ref = repo.head.ref
+            assert hasattr(ref, "name")
+            assert isinstance(ref.name, str)
+            if head_ref_name := ref.name:
```
awelzel: It seems `head_ref_name` is superflous:
```
if ref.name:
    if ref.name.startswith("origin/"):
        return ref.name.split("/", 1)[-1]
    return ref.name
```

zeekpkg/_util.py
```diff
@@ -290,7 +305,7 @@ def git_remote_urls(repo):
     return remotes
 
 
-def is_sha1(s):
+def is_sha1(s: str | None) -> bool:
```
awelzel: Can we force this to just `str`? Seems reasonable expect a caller to do the None check.

zeekpkg/manager.py
```diff
@@ -446,7 +451,7 @@ def _write_plugin_magic(self, ipkg):
                             exception,
                         )
 
-    def _read_manifest(self):
+    def _read_manifest(self) -> tuple[str, str, str]:
```
awelzel, line 454: Mind updating the doc string to match this?

zeekpkg/manager.py
```diff
@@ -2722,18 +2832,22 @@ def _stage(self, package, version, clone, stage, env=None):
...
+        invalid_reason: str | None = _parse_package_metadata(
+            metadata_parser,
+            metadata_file,
+        )
         if invalid_reason:
             return invalid_reason
 
         metadata = _get_package_metadata(metadata_parser)
-        metadata, invalid_reason = self._interpolate_package_metadata(metadata, stage)
+        metadata2, invalid_reason = self._interpolate_package_metadata(metadata, stage)
```
awelzel: Mind calling this `interpolated_metadata` or `metadata_interp` or is that too long?

zeekpkg/manager.py
```diff
@@ -687,16 +692,16 @@ def find_builtin_package(self, pkg_path):
...
 
         return None
 
-    def installed_packages(self):
+    def installed_packages(self) -> list[InstalledPackage]:
         """Return list of :class:`.package.InstalledPackage`."""
         return [ipkg for _, ipkg in sorted(self.installed_pkgs.items())]
 
-    def installed_package_dependencies(self):
+    def installed_package_dependencies(self) -> dict[str, dict[str, str] | None]:
```
awelzel, line 704: Could we make `ipkg.package.dependencies()` to not return `None` and the one case where it does `raise ValueError` or an Assertion instead?

I understand you didn't want to modify the API, but dependencies should either exist or be empty. Even having `None` is just super weird, especially if both callers seem to ignore `None` anyhow.

awelzel (approved):
Approving, but I'm pretty partial here, so if this helps `zkg`, :ship: and :man_shrugging:

zeekpkg/_util.py
bbannier: Yepp, the assertions guide mypy so it can remove `None` branches.

zeekpkg/manager.py
bbannier: The issue is that validation isn't well encapsulated and many of these functions just take a naked `metadata: dict[str, str]` which leads to functions doing their own ad hoc validation which is not great (`user_vars` just below has the same pattern). TBH, I'd prefer not trying to clean this up right now.

awelzel: yeah, works for me, I guess it's clear that `dict[str, dict[str, str] | None]` for dependencies is a bit strange :-)

bbannier: Agree 100%. I'd argue that's the kind of cruft one gets when developing without type hints ;)

PR #210, opened 2025-08-17
Fix typing of use of `readline.set_startup_hook`

bbannier:


zeekpkg/uservar.py
```diff
@@ -137,7 +138,6 @@ def resolve(
 
         desc = " (" + self._desc + ")" if self._desc else ""
         print(f'"{name}" requires a "{self._name}" value{desc}: ')
-        assert val
         self._val = _rlinput(self._name + ": ", val)
```
awelzel: I'd probably do something like this instead and leave the typing of `_rlinput()` alone. A function having a default value of `""`, but then also accepting a `None` even though `""` and `None` are treated the same tickles "this is strange".

```suggestion
        self._val = _rlinput(self._name + ": ", prefill=val if val else "")
```

But don't think that matters enough really.

awelzel (approved):
LGTM :rocket:

An explicit sentence that this was broken by #209 would've been good to help context switch, but I've stitched it together that one of the added `assert`s was too optimistic :+1:

zeekpkg/uservar.py
bbannier: Good idea.

TBH, I was slightly shocked how little test coverage we had since the reproducer was so basic. Let's hope we can shake out any remaining issues before 8.1.

PR #180, opened 2024-02-18
Revert "Remove references to bro-pkg.meta and bro-pkg.index"

timwoj:

PR #212, opened 2025-09-10
Touchup some docs

evantypanski:
I was looking at the `zkg` docs to make a section in a Zeek tutorial and noticed that they need a little love. Here's part of that love.

I thought about enabling some of the RST lints in the Zeek docs, like double-backticks for code, but apparently it's in the [docs style guide](https://docs.zeek.org/projects/package-manager/en/stable/developers.html#style-conventions) for ZKG to use single backticks sometimes. And I don't feel like walking through 50 cases of that right now to clean it up, these were just random things I noticed reading through

### DIFF
--- a/doc/package.rst
+++ b/doc/package.rst
@@ -22,12 +22,6 @@ This is the package's metadata file in INI file format and may contain
 :ref:`additional fields <metadata-fields>` that describe the package as well
 as how it inter-operates with Zeek, the package manager, or other packages.
 
-.. note::
-
-   :file:`zkg.meta` is the canonical metadata file name used since
-   :program:`zkg` v2.0.  The previous metadata file name of :file:`bro-pkg.meta`
-   is also accepted when no :file:`zkg.meta` exists.
-
 .. _package-shorthand-name:
 
 Note that the shorthand name for your package that may be used by :ref:`zkg
--- a/doc/package.rst
+++ b/doc/package.rst
@@ -24,9 +24,9 @@ as how it inter-operates with Zeek, the package manager, or other packages.
 
 .. note::
 
-   :file:`zkg.meta` is the canonical metadata file name used :program:`since
-   zkg v2.0`.  The previous metadata file name of :file:`bro-pkg.meta` is also
-   accepted when no :file:`zkg.meta` exists.
+   :file:`zkg.meta` is the canonical metadata file name used since
+   :program:`zkg` v2.0.  The previous metadata file name of :file:`bro-pkg.meta`
+   is also accepted when no :file:`zkg.meta` exists.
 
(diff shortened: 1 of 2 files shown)


evantypanski:
Sorry I was silly and included something I shouldn't; removed and put them in the same small commit

doc/package.rst
```diff
@@ -24,9 +24,9 @@ as how it inter-operates with Zeek, the package manager, or other packages.
 
 .. note::
 
-   :file:`zkg.meta` is the canonical metadata file name used :program:`since
-   zkg v2.0`.  The previous metadata file name of :file:`bro-pkg.meta` is also
-   accepted when no :file:`zkg.meta` exists.
+   :file:`zkg.meta` is the canonical metadata file name used since
+   :program:`zkg` v2.0.  The previous metadata file name of :file:`bro-pkg.meta`
+   is also accepted when no :file:`zkg.meta` exists.
```
timwoj: Should we mention that bro-pkg.meta is deprecated and can't be depended on to work forever? Even though it will probably work forever because we can't update all of the packages using it?

timwoj (approved):

doc/package.rst
evantypanski: Hm, I didn't just because we will warn about it (from b5e089d25ea59fc2a0cb96cafcd1caf5b1eab4a1)

This seems more like a statement of information. What do you think about just removing this note here? Other places will still mention it's "legacy" and warn, but this note would almost always get ignored, and right now it's very prominent..

EDIT: BTW That commit says remove in 3.0 ¯\\\_(ツ)_/¯

timwoj: Yeah, I tried to remove all of the bro things entirely from zkg once. It didn't work out and we reverted a good chunk of it. Arne's comment on https://github.com/zeek/package-manager/pull/180 mentions removing references from the docs though, so maybe this is an instance that can go.

Or maybe change it to something like "bro-pkg.meta is also accepted, but is only kept as valid for existing legacy packages that use it. Any new packages should use zkg.meta".

evantypanski: > so maybe this is an instance that can go.

I'm leaning towards that (hence the commit I added that does it :) ) because this is a "how-to" rather than explaining general concepts. Before this in the `zkg` section it already mentions `bro-pkg.meta`, so it's not like this is the only reference to the legacy file. But since this is explaining how to create a package, I don't even think the mention belongs there at all. The big note just feels out of place for something 6 years old in a presumably current tutorial

timwoj (approved):

doc/package.rst
timwoj: Works for me!

PR #218, opened 2026-01-08
Do not require that user vars are non-empty

bbannier:
When typing this code in #209 we introduced checks that variables of types like `str | None` were set. In making that we accidentially used a narrower requirement that user vars were not only set, but also not empty. This is because

    assert val

for a `val: str` means "is set, and its value is truthy (i.e., the `str` is not empty).

Since we still catch empty users vars elsewhere this still had the same effect, but now potentially exposed users to a naked `AssertionError` instead of an error message.

With this patch we just require that user variables are set.


bbannier:
@timwoj This was a regression in the window between zeek-8.0 and zeek-8.1-rc2, we could think about including this in zeek-8.1.0, or in its first patch release. The change here seems pretty narrow to me.

awelzel (approved):
A test would've been nice, but I guess that works.

PR #223, opened 2026-06-15
Add pytest wrapper around btest and collect coverage information

### LINKED ISSUE #221: Collect coverage during testing
We should explore collecting coverage during testing so we have an understand what code is actually exercised. Currently the test suite uses `btest`, so we cannot directly hook into tools like `pytest-cov`, but there are probably around this, e.g., migrating the complete test suite to `pytest`, by invoking `btest` in some clever way, or by collecting coverage information in some other way.

bbannier:


awelzel (approved):
Chatted with Benjamin in a 1:1 about this and LGTM.

Only thought is whether the environment variables for coverage should be passed explicitly to `subprocess.run()` instead of globally in `conftest.py` and assumed to be inherited, but no strong opinion.

bbannier:
> Only thought is whether the environment variables for coverage should be passed explicitly to `subprocess.run()` instead of globally in `conftest.py` and assumed to be inherited, but no strong opinion.

While coverage should be automatically collected for unit tests (test code directly importing the `zeekpkg` module), tests launching e.g., `zkg` need this setup. Moving this to just the BTest invocation would mean coverage would not be collected if we ever add unit test-style pytests launching `zkg` without BTest. Not the case right now, but what we have here should be more futureproof.

PR #226, opened 2026-06-19
Clean up Python dependency management 

bbannier:


doc/developers.rst
```diff
@@ -42,13 +42,13 @@ Dependencies
...
 To build documentation locally, find the requirements in
-:file:`requirements.txt`:
-
-  .. literalinclude:: ../requirements.txt
+:file:`pyproject.toml`:
 
 They can be installed like::
 
-  pip3 install -r requirements.txt
+  python3 -m venv .venv
```
ckreibich, line 49: Do I _have_ to use a virtualenv to be able to pip-install the packages? I ask because while I appreciate virtualenvs and use them often, I also try to maintain a "normal" installation of the dependencies, and on Linux the first step for that is to check the system packages, and then fall back to systemwide regular pip install.

ckreibich (commented):
Nice, I'm on board with all of these. The only thing I wonder about is the first commit, i.e.  whether we should still keep an explanation for the needed dependencies if for some reason the installation is messed up or non-standard (say because the user wants to rely on system packages to provide them but something goes wrong — similar to my comment over on #228).

The help output could be more streamlined — a simple pointer to the docs perhaps — but would a bit more graceful than an unhandled `ImportError`.

doc/developers.rst
bbannier: No, you do not _have to_. The `pyproject.toml` equivalent of `pip install -r requirements.txt` would be `pip install -e .` (the `-e` ensures that edits in the source files have an effect during development). If you run this outside of a virtualenv the dependencies of the package would be installed into the system, potentially breaking any other scripts which directly use system packages (e.g., due to version conflicts).

A slightly less extreme approach which still allows you to pull in unspecified system packages would be to give the virtualenv access to system packages by setting it up with

```console
$ python3 -m venv --system-site-packages .venv
```

This sets up the virtualenv so packages from the system's `site-packages` are immediately available as well. Since these dependencies are not expressed, the enviroment is not reproducible.

I'd like to avoid documenting any of that, and stick with the more standard approach I proposed in this PR. IMO installing deps into the system or `--system-site-packages` is a recipe for hard to diagnose failure modes, and does not enable anything not already possible in a well-specified setup using dependencies expressed in `pyproject.toml`. I'd also argue that installing Python dependencies into the system has not been normal since at least python-3.0 in 2008.

bbannier:
> The only thing I wonder about is the first commit, i.e. whether we should still keep an explanation for the needed dependencies if for some reason the installation is messed up or non-standard (say because the user wants to rely on system packages to provide them but something goes wrong — similar to my comment over on #228).

I have never seen this anywhere else, and it seems weird. Unless looking at very simple scripts without any dependencies, I'd expect all Python projects to have some `pyproject.toml` which spells out how the package can be installed.

doc/developers.rst
ckreibich: Got it. It's fine with me, we'll need to make a pretty big docs pass eventually anyway and can finetune then. I just wanted to understand what's required vs common usecase.

ckreibich (approved):

PR #227, opened 2026-06-22
Stop emitting `[template]` and `[template_vars]` sections in `zkg.meta`

### LINKED ISSUE #220: package-template `zkg.meta` [template] context handling
The [`package-template` README](https://github.com/zeek/package-template#status) notes that template support is 

> "a beta feature and some functionality is still undergoing changes. Early feedback, feature requests, and bug reports are all very welcome." 

In that spirit, some feedback from getting to grips with Spicy:

I followed developer guidance flagging `spicy-tftp` as a good starting point, and noticed the `[template]` block in [`zkg.meta`](https://github.com/zeek/spicy-tftp/blob/main/zkg.meta) show a branch that no longer exists:

```ini
[template]
source = ssh://git@github.com/zeek/package-template
version = topic/robin/spicy-updates
commit = 990dde20
```

As far as I can tell this is harmless: the `version`/`commit` lines look like a record of which template revision generated the package at `zkg create` time, not a live dependency.

So this is really two small questions rather than a bug:
- Worth a doc note clarifying that `[template] version`/`commit` are a creation-time stamp, not a tracked ref? ~~It reads like something `zkg` keeps current~~, which is what sent me looking.
- Are said fields even needed? If nothing consumes them, a dangling value can be confusing.

*Minor aside*
- while comparing notes I noticed other Spicy packages stamp `[template]` a few different ways (and one not at all), which is partly what made me unsure:

| Package | `source` | `version` | `commit` |
|---------|----------|-----------|----------|
| [spicy-tftp](https://github.com/zeek/spicy-tftp/blob/main/zkg.meta) | `zeek/package-template` | `topic/robin/spicy-updates` (deleted) | `990dde20` |
| [spicy-http](https://github.com/zeek/spicy-http/blob/main/zkg.meta) | `bbannier/package-template-spicy` | `master` | `c14f0756` |
| [spicy-ldap](https://github.com/zeek/spicy-ldap/blob/main/zkg.meta) | `package-template-spicy` | `master` | — |
| [spicy-dns](https://github.com/zeek/spicy-dns/blob/main/zkg.meta) | — | — | — |

---

> **AI/LLM disclosure** — per [community guidelines on AI-assisted contributions](https://community.zeek.org/t/zeek-newsletter-issue-60-february-2026/7953), I used an LLM to help investigate and draft this report; the key findings and any code change were human-verified.

bbannier:
These sections were always informational, and we never implemented the functionality to update the instantiated template on template updates. It is unlikely that we will ever be able to reliably implement this with the current loosly defined template. The sections were also potentially confusing to whoever looked at them since they referred to the state at template instantiation, but never reflected subsequent edits.

This commit drops them without replacement.

Closes #220.

### DIFF
--- a/doc/package.rst
+++ b/doc/package.rst
@@ -102,27 +102,7 @@ existing package. When the requested output directory exists, it will
 prompt for permission to delete the existing directory.
 
 After instantiation, the package is immediately installable via
-:program:`zkg`. You'll see details of how it got generated in its
-initial commit, and the newly minted ``zkg.meta`` has details of the
-provided user variables:
-
-.. code-block:: console
-
-    $ cat foobar/zkg.meta
-    ...
-    [template]
-    source = package-template
-    version = master
-    zkg_version = 2.8.0
-    features = plugin
-
-    [template_vars]
-    name = Foobar
-    namespace = MyOrg
-
-This information is currently informational only, but in the future
-will enable baselining changes in package templates to assist with
-package modernization.
+:program:`zkg`.
 
 To keep templates in sync with :program:`zkg` versions, templates
 employ semantic API versioning. An incompatible template will refuse
--- a/testing/baselines/tests.template-create/out3.zkg.meta
+++ b/testing/baselines/tests.template-create/out3.zkg.meta
@@ -3,14 +3,4 @@
 script_dir = scripts
 summary = TODO: A summary of test3 in one line
 description = TODO: A more detailed description of test3.
-	It can span multiple lines, with this indentation.
-
-[template]
-source = foo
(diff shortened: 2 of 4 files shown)


bbannier:
@ckreibich How does this look to you? It seems the half-feature we have ATM could be more confusing than useful, and I am not aware where what we have has ever helped use debug issues.

ckreibich (approved):
Yeah, I'm on board. I think it's quite unlikely that we'll implement some sort of tooling based on this information at this point, and the version-use context is also in the initial commit generated by the templating. We should keep that one, and your PR doesn't change that.

Issue #202, opened 2025-01-28
Add argument for cleaning scratch directory after installation


timwoj:
This came up as part of the [rework of zeek-pkg-web](https://github.com/zeek/zeek-pkg-web/pull/7). It would be nice if `zkg` had an option to tell it to delete the files downloaded in to the scratch directory after it is done installing/checking/etc a package. In the context of zeek-pkg-web, we download the package to run the `info` command but then leave the scratch directory behind once done.

bbannier:
Is this a subtask or plain dup of 177?

timwoj:
> Is this a subtask or plain dup of [#177](https://github.com/zeek/package-manager/issues/177)?

I could go either way. Having it do it automatically without having to call a second command would be nice.

Issue #177, opened 2024-01-23
Clean up `scratch` space and provide cleanup helper


ckreibich:
`zkg` generally does not clean up after itself when it puts stuff in its `scratch` directory. For users going through a lot of packages this can add up (since `zkg` may place a clone for each package in the scratch space). It would also be nice to add a cleanup function to tidy on demand, a bit like `dnf clean` / `apt clean`.

ottobackwards:
How would one know what to clean?  can this be expanded to have that information?
Or is it just scratch?

bbannier:
AFAICT the following directories under zkg's statedir do not contain information required at runtime:

- `clones`
- `logs`
- `scratch`
- `testing`

Even compiled plugins should be contained in `plugin_dir` so that removing any of the above dirs shouldn't e.g., cause debug symbols to become unavailable.

We might want to have separate cleanups for each directory, e.g.,

```
zkg prune [ --clones | --logs | --scratch | --testing | --all ]
```

ottobackwards:
Is rebuild a consideration?  Did that ever make it in?

ottobackwards:
Would a good scope limitation be splitting automatic cleanup and explict commands into two different issues?

Also, would the commands be as @bbannier listed, for ALL things or for specific plugins?
IE>
`zig prune [ --clones | --logs | --scratch | --testing  | --all ] [plugin]`

Would it be plugin@version? or the same specifier acceptance and the install command?

bbannier:
> Is rebuild a consideration? Did that ever make it in?

That's #38.

> Would it be plugin@version? or the same specifier acceptance and the install command?

IMO these should be global, not package-specific (mirroring what e.g., `apt`/`dnf` mentioned in the original issue do), but maybe @ckreibich has other ideas.

ottobackwards:
@ckreibich ?

PR #234, opened 2026-07-03
Code reorg

ckreibich:
@bbannier this is the bulk of the changes I've been sitting on that we've discussed. In essence it's moving all of the commands out of zkg into zeekpkg.cli, the introduction of a config class, a UI abstraction, and some minor tweaks.

Some minor behavior changes slipped in there that don't need to be there, such as 04a8580d752c6a388804f2bef93a54e255df63b4. We could postpone these until later — I don't feel strongly.

I'll make this a draft PR since I imagine you have opinions about it, and I don't mean for this to cause conflicts with other PRs in flight.


zeekpkg/config.py
```diff
@@ -0,0 +1,369 @@
...
+        """
+        return os.path.join(self.plugin_dir(), "packages")
+
+    def sources_clone_dir(self) -> str:
+        """Directory where zkg clones package "source" repos.
+
+        These are zkg's package inventories. Each source gets a subdirectory
+        associated with its name. The default source is called "zeek".
+        """
+        return os.path.join(self.state_dir(), "clones", "sources")
```
bbannier, line 275: This was called `clones/source` previously and since we don't have migration logic in place this will silently break existing zkg installations. I understand that the renaming makes things more tidy, but I think we should either avoid breaking users for such a minor improvement and skip this rename, or add migration logic. The first would probably be simpler.

zeekpkg/config.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def packages_clone_dir(self) -> str:
+        """Directory where zkg keeps the source tree of installed packages.
+
+        Each package gets a subdirectory associated with its name.
+        """
+        return os.path.join(self.state_dir(), "clones", "packages")
```
bbannier, line 282: Same here.

zeekpkg/logs.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    handler = logging.StreamHandler()
+    handler.setFormatter(formatter)
+
+    if verbosity == 1:
+        LOG.setLevel(logging.INFO)
+    elif verbosity == 2:
+        LOG.setLevel(logging.WARNING)
+    elif verbosity >= 3:
```
bbannier, line 30: This is weird, I'd expect `WARNING` to be less verbose than `INFO`. Did you accidentally swap these or did I miss something?

zeekpkg/manager.py
```diff
@@ -3245,11 +3042,13 @@ def _interpolate_package_metadata(
             return None, "package has malformed 'user_vars' metadata field"
 
         substitutions = {
-            "zeek_dist": self.zeek_dist,
             "package_base": stage.clone_dir,
         }
 
-        substitutions.update(self.user_vars)
+        if CONFIG.zeek_dist() is not None:
```
bbannier, line 3048: `Config.zeek_dist` returns a `str` and implements a fallback so it currently can never return `None` and this seems unreachable; there's also a `XXX` TODO in the code there. Suggest to remove the fallback and clearly communicate the unset value up by returning `str | None`. Only issue is that our fallback here then uses `Config.zeek_dist` which also returns a `str` and has a similar fallback, so we might need to deal with unset values somewhere else completely anyway.

zeekpkg/config.py
```diff
@@ -0,0 +1,369 @@
...
+
+        self.canonicalize()
+        self.sanity_check()
+
+    # def items_without_defaults(section: str) -> list[tuple[str, str]]:
+    #     # Same as ConfigParser.items(section), but without default keys.
+    #     defaults = {key for key, _ in self.items("DEFAULT")}
+    #     items = sorted(self.items(section))
+    #     return [(key, val) for (key, val) in items if key not in defaults]
+
```
bbannier, line 144: Dead code?

zeekpkg/manager.py
```diff
@@ -725,15 +663,27 @@ def loaded_packages(self) -> list[InstalledPackage]:
...
                 a package source called "zeek" with package named "foo" in
                 :file:`alice/zkg.index`, the following inputs may refer
                 to the package: "foo", "alice/foo", or "zeek/alice/foo".
+
         """
         name = name_from_path(pkg_path)
-        return os.path.join(self.log_dir, f"{name}-build.log")
+        return os.path.join(CONFIG.log_dir(), f"{name}-build.log")
+
+    def package_test_log(self, pkg_path: str) -> str:
```
bbannier, line 684: This should have a docstring like the similar `package_build_log` you also added.

zeekpkg/ui.py
```diff
@@ -0,0 +1,335 @@
...
+    This allows us to use a toplevel UI object whose implementation we can swap out,
+    keeping the global UI working in other modules that imported from this module.
+    A toplevel assignment would break the relationship.
+    """
+
+    def __init__(self) -> None:
+        self.impl: UserInterface = PlainUI()
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.impl, name)
```
bbannier, line 332: This is pretty magical and also completely disables any type checking since to support all method signatures we need to return `Any`. Is there a reason we couldn't instead just subclass `UserInterface` and implement its interface instead of this wildcard approach?

zeekpkg/cli.py
```diff
@@ -0,0 +1,2707 @@
...
+
+    prompt_for_user_vars(
+        manager,
+        args,
+        [info for info, _, _ in package_infos],
+    )
+
+    # Run tests on explicitly requested packages. With --force, this
+    # exits with error if tests fail.
+    # XXX this prints output even with --force
```
bbannier, line 321: Wasn't that the case even previously?

zeekpkg/manager.py
```diff
@@ -587,6 +520,8 @@ def add_source(self, name: str, git_url: str) -> str:
                 version=version,
             )
         except git.GitCommandError as error:
+            # XXX seems this could also error when requesting nonexisting
+            # branches/versions?
```
bbannier, line 524: Probably, but even before your changes, right? Let's either create an issue, or drop this, tending to the latter.

zeekpkg/manager.py
```diff
@@ -748,6 +698,8 @@ def match_source_packages(self, pkg_path: str) -> list[Package]:
         rval = []
         canon_url = canonical_url(pkg_path)
 
+        # XXX This is the right place to populate a package source on-demand.
+
```
bbannier, line 702: Probably, but not needed here. Let's just drop this.

zeekpkg/manager.py
```diff
@@ -798,7 +750,11 @@ def has_scripts(self, installed_pkg: InstalledPackage) -> bool:
         Returns:
             bool: True if the package has installed Zeek scripts.
         """
-        return os.path.exists(os.path.join(self.script_dir, installed_pkg.package.name))
+        # XXX is this better than relying on the presence of the scripts
+        # directive in zkg.meta?
```
bbannier, line 754: Not needed here.

zeekpkg/manager.py
```diff
@@ -810,13 +766,18 @@ def has_plugin(self, installed_pkg: InstalledPackage) -> bool:
         Returns:
             bool: True if the package has installed a Zeek plugin.
         """
-        return os.path.exists(os.path.join(self.plugin_dir, installed_pkg.package.name))
+        # XXX is this better than relying on the presence of the build_command
+        # in zkg.meta?
```
bbannier, line 770: Same.

zeekpkg/manager.py
```diff
@@ -851,11 +811,14 @@ def save_temporary_config_files(
...
 
-            backup_file = os.path.join(self.scratch_dir, "tmpcfg", config_file)
+            backup_file = os.path.join(CONFIG.scratch_dir(), "tmpcfg", config_file)
             make_dir(os.path.dirname(backup_file))
             shutil.copy2(config_file_path, backup_file)
             rval.append((config_file, backup_file))
 
+        # XXX there's nothing temporary about the files just created, and
+        # nothing that cleans them.
+
```
bbannier, line 821: Should this be in the docstring, or can we drop this?

zeekpkg/manager.py
```diff
@@ -968,54 +936,38 @@ def backup_modified_files(
...
+            )
             timestamp = time.strftime(".%Y-%m-%d-%H:%M:%S")
             backup_path = os.path.join(backup_dir, filename + timestamp)
             make_dir(backup_dir)
             shutil.copy2(install_path, backup_path)
             rval.append(backup_path)
 
-        return rval
+        # XXX odd to have both this method and save_temporary_config_files(),
+        # since the latter only copies what's in the clone tree?
```
bbannier, line 951: Did you want to implement a cleanup? Let's not leave the TODO in any case.

zeekpkg/manager.py
```diff
@@ -968,54 +936,38 @@ def backup_modified_files(
...
-
-            package_issues (list of (str, str)): a list of reasons for
-                failing to collect metadata per packages/repository.
-                The first tuple element gives the repository URL in which
-                the problem occurred and the second tuple element describes
-                the failure.
-        """
+        # XXX also note that there's no restore_modified_files() because
+        # restoration isn't the point of these backups, it's to preserve the
+        # modified version for reference/review (similar to e.g. .rpmsave).
```
bbannier, line 956: Same

zeekpkg/manager.py
```diff
@@ -1359,13 +1102,19 @@ def upgrade(self, pkg_path: str) -> str:
         if ipkg.status.tracking_method == TRACKING_METHOD_COMMIT:
             # The above check for whether the installed package is outdated
             # also should have already caught this situation.
+
+            # XXX This likely refers to the fact that if a package got installed
+            # to a given commit, there's only that single version, so it cannot
+            # be outdated. So this is effectively pins the package.
```
bbannier, line 1108: Let's either incorporate this in the branch comment above, or drop it.

zeekpkg/manager.py
```diff
@@ -2920,12 +2712,16 @@ def _stage(
         script_dir_src = os.path.join(clone.working_dir, pkg_script_dir)
         script_dir_dst = os.path.join(stage.script_dir, package.name)
 
+        # XXX what if the package has only a plugin, no script_dir?
+        # Seems conceivable and not an error.
```
bbannier, line 2716: Let's either fix this with an additional commit, or capture it in an issue.

zeekpkg/manager.py
```diff
@@ -3217,6 +3011,9 @@ def _install(
         if not package.source:
             # If installing directly from git URL, see if it actually is found
             # in a package source and fill in those details.
+            #
+            # XXX This seems strange -- what if this is ambiguous across
+            # sources?  Also we overwrite the metdata below anyway.
```
bbannier, line 3016: New issue?

zeekpkg/manager.py
```diff
@@ -3364,8 +3162,12 @@ def _copy_package_dir(
     if not os.path.exists(src):
         return ""
 
+    # XXX instead of magic handling of tarfiles, it'd be nicer
+    # to have separate keywords for shipping archives.
+    # Since it only seems to make sense for pre-built plugins
+    # and their dependencies, perhaps plugin_archive.
```
bbannier, line 3168: I suspect this could be implemented on top of https://github.com/zeek/package-manager/pull/236, maybe let's add a comment there and I can work on incorporating it, or let's add an issue sketching what you had in mind.

zeekpkg/logs.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def configure(verbosity: int = 0) -> None:
+    """Configures logging.
+
+    Args:
+        verbosity (int): the log level. Values 0-3 increase logging, larger
+            values make no difference.
+
+        rich_logging (bool): whether to use timestamped, log-style log formatting.
```
bbannier, line 16: We don't have this parameter, is this stale or did you want to add it?

zeekpkg/config.py
```diff
@@ -0,0 +1,368 @@
...
+)
+from .consts import (
+    ZEEK_ZKG_CONFIG_DIR,
+    ZEEK_ZKG_STATE_DIR,
+    ZKG_DEFAULT_SOURCE,
+    ZKG_DEFAULT_TEMPLATE,
+)
+
+
+class Config(configparser.ConfigParser):
```
bbannier, line 25: Could you add a class docstring and also ones for the interesting methods? Right now this is pretty bare-bones.

zeekpkg/ui.py
bbannier, line 1: Let's add a module-level docstring here.

zeekpkg/cli.py
```diff
@@ -0,0 +1,2707 @@
...
+def cmd_info(
+    manager: Manager,
+    args: argparse.Namespace,
+) -> None:
+    if args.version and len(args.package) > 1:
+        UI.error('"info --version" may only be used for a single package')
+        sys.exit(1)
+
+    # Dictionary for storing package info to output as JSON
+    pkginfo: dict[str, Any] = {}
```
bbannier, line 1606: Just a note, nothing to do: You just moved this of `Any` around (same for the use in `_fill_metadata_version` below), but we should remember to use proper types for new code instead of sidestepping typechecking with `Any`.

One way to do that would be to introduce proper types like e.g.,

```python
PkgMetadataVersion = dict[str, str | dict[str, str]]


class PkgInfoEntry(TypedDict, total=False):
    metadata: dict[str, PkgMetadataVersion]
    invalid: str
    url: str
    versions: list[str]
    install_status: dict[str, str]
    metadata_file: str
```

With that we'd have a `pkginfo: dict[str, PkgInfoEntry] = {}` here, and pass `PkgMetadataVersion` to `_fill_metadata_version` instead of `Any`. Our currently handling is inconsistent, so we'd need to e.g., add some assertions to make clear what exact shape the data we work with has. But in any case, no need to worry about that in this branch.

zeekpkg/config.py
ckreibich: Sounds good to me, Benjamin. It's true that in a few places this isn't purely a code reorg. I'm fine with separating such changes out. Fwiw, I assumed form the get-go that the next version would not be backward compatible.

zeekpkg/config.py
ckreibich: (yep)

zeekpkg/logs.py
ckreibich: You're right, that's wrong. It's weird because I remember getting confused about what we're logging at which level, but I can no longer track it down. It's possible that I stumbled over the above, i.e. my own bug. One way or another, it's definitely wrong — thanks for catching this.

zeekpkg/config.py
ckreibich: Fwiw there's a similar change with `clones/package` -> `clones/packages`, from 86c1cbf14e8fd597418bb52cd66db50cb5e929ae.

zeekpkg/config.py
ckreibich: Yeah, and I don't recall why it's there. Some of that config code is taken from zeek-client, but I don't see it in there either. It can go.

zeekpkg/ui.py
ckreibich: If you see an approach that works, go for it. I think the problem I faced was that if multple modules use `UI.foo()`, and this module reassigns the `UI` global to a different implementation, that change does not propagate to the other modules. By instead reassigning the `impl` in the proxy instance, it does. I'm sure there are better ways to do this — I considered this more of a workaround and for sure don't feel strongly. (I did consider a function, but I didn't want to say `UI().foo()` if we have `CONFIG.foo`, `LOG.foo` etc for the other singletons.)

zeekpkg/cli.py
ckreibich: I think that referred to the fact that `--force` currently has pretty mixed semantics and in some cases output occurs only without `--force`, i.e., it also acts as strange `--quiet`. We should definitely clean this up and distinguish output verbosity from the degree to which it needs input.

zeekpkg/manager.py
ckreibich: Sounds good.

zeekpkg/manager.py
ckreibich: It looks to me like #177 / #202 will take care of this one.

Fwiw, I do think we should have the freedom to annotate our own code with things that aren't quite right or could be better. Those aren't necessarily bugs, and they don't all rise to needing tickets. If this is just about the literal "XXX", I could just see dropping that in some of those comments.

zeekpkg/manager.py
ckreibich: But this isn't a TODO ... it's just a comment on the code. I don't see a better way to comment things that we won't do right away, that aren't as localized in the git commit message, and don't need a ticket. I basically don't know what to do with this stuff right now.

Here too, if the "XXX" is the trigger, just remove that.

bbannier: Like I mentioned offline, they looked leftover work-in-progress notes to self from you. If you want to keep them in let's simply make them proper comments, there shouldn't be a need for that magic `XXX` marker.

ckreibich:
Okay Benjamin, I'm getting on this one now — will work in edits and rebase, then take out of draft.

ckreibich:
Alright! This is a full rebase and things seem to be working pretty well. This still has minor feature changes, which I think we should actually aim to keep for Zeek 9 (particularly aligning the handling of test and build logs, putting them in one place). But we can take that out of this one for subsequent PRs, doesn't matter to me.

ckreibich:
Btw I don't feel strongly about whether this should make Zeek 9 still. It's a big change to mainly put in place cleanups. I could also see shipping what's currently in master with 9 (so your new bundling approach etc), and considering this PR the starting point for all of the upcoming work.

zeekpkg/logs.py
ckreibich: Nice catch. I removed it in the latest version, so the timestamped output is now the norm. I think that's a good thing, since (in addition to going to stderr)  it helps distinguish this output from the general UI-level output.

zeekpkg/logs.py
ckreibich: Fixed in the latest version. I'm thinking we should go more fine-grained in a future revision, not logging anything by default (other than perhaps CRITICAL, which currently isn't used).

zeekpkg/ui.py
ckreibich: Yep, done.

ckreibich:
All yours Benjamin. I missed one change in our conversation today, 40f995d8a796412727b4fc7ec0cdb13b0e937f82. It'd be nice to have because it's long been too onerous to locate test log output, but we can take it out of this PR.

If I've not addressed any of your comments at this point, that's an oversight.

zeekpkg/manager.py
bbannier: You dropped the `XXX`, but still kept it in. IMO this should either be a `NOTE` explaining current behavior, or a `TODO` capturing possible future improvements. This comment here reads more like a slightly confusing/confused musing which I don't think is that useful and could be dropped (to be frank, I feel this makes it harder to get a feel for the code). Since you want to keep this I'll resolve it though 🤷.

zeekpkg/manager.py
bbannier: Same argument.

zeekpkg/config.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        for key, path in self.items("paths"):
+            path = os.path.expanduser(path)  # Expand ~
+            path = os.path.expandvars(path)  # Env var substitution
+            if path:
+                # Canonicalize; no trailing slashes -- but only
+                # if there is a value, because normpath otherwise
+                # produces "."
+                path = os.path.normpath(path)
+                self.set("paths", key, path)
```
bbannier, line 113: [Previously](https://github.com/zeek/package-manager/blob/35b4461e32c24a000216bc57ea8aad5b135fa0d3/zkg#L261-L268) we would also expand `~` and env vars in the `sources` and `templates` and we now silently ignore them; this could break existing configs.

```suggestion
        for section in ("paths", "sources", "templates"):
            for key, path in self.items(section):
                path = os.path.expanduser(path)  # Expand ~
                path = os.path.expandvars(path)  # Env var substitution
                if path:
                    # Canonicalize; no trailing slashes -- but only
                    # if there is a value, because normpath otherwise
                    # produces "."
                    path = os.path.normpath(path)
                    self.set("paths", key, path)
```

zeekpkg/config.py
```diff
@@ -0,0 +1,382 @@
...
+                    "state_dir": state_dir,
+                },
+                "sources": {},
+                "templates": {},
+            },
+        )
+
+        # These may turn out to be empty to populate only if present:
+        if val := os.getenv("ZKG_DEFAULT_SOURCE", ZKG_DEFAULT_SOURCE):
+            self.set("sources", "zeek", val)
```
bbannier, line 63: This now always injects the default source even into an existing config while [previously](https://github.com/zeek/package-manager/blob/35b4461e32c24a000216bc57ea8aad5b135fa0d3/zkg#L203-L206) we'd only inject if no config file (which would imply: no other source) was present. This seems to mean that users cannot opt out of the zeek source anymore which would be a pretty bad regression.

The issue is that we do not know here whether we have a `configfile` anymore. Since the default for `templates` is a proper default config it makes sense to keep this here, but we'd only want to set `sources` if there was no config so that part could and IMO should be moved to the caller after their `reset()` call. Does that make sense?

zkg
```diff
@@ -3,3179 +3,33 @@
...
 
-            if extdeps:
-                extdep_listing += f"  from {name} ({version}):\n"
+    zeekpkg.CONFIG.reset(args)
+    zeekpkg.CONFIG.update_from_file(configfile)
+    zeekpkg.CONFIG.update_from_args(args)
 
-                for extdep, semver in sorted(extdeps.items()):
-                    extdep_listing += f"    {extdep} {semver}\n"
+    manager = zeekpkg.manager.Manager()
```
bbannier: We'd [previously](https://github.com/zeek/package-manager/blob/35b4461e32c24a000216bc57ea8aad5b135fa0d3/zkg#L343-L390) catch exceptions of this constructor, e.g., `OSError` and show users a more useful error message, but we lost that and now would simply let the interpreter dump a stacktrace. This could e.g., show up if the state_dir is owned by `root`, and a user tries to run an operation like `zkg install ...`. Could we get that better error message back?

zeekpkg/config.py
```diff
@@ -0,0 +1,382 @@
...
+        args directly include a config file (`--configfile`), it returns its
+        path. Second, if `--user` is given, it attempts to locate the config
+        file in the user's home. Third, it attempts the ZKG_CONFIG_FILE
+        environment variable, and finally, the default configuration directory.
+
+        Returns the path to the first config file successfully located, and the
+        empty string if all attempts fail.
+        """
+        if args.configfile and file_is_not_empty(args.configfile):
+            return str(args.configfile)
```
bbannier, line 372: We previously had a check for non-exisiting config files [here](https://github.com/zeek/package-manager/blob/35b4461e32c24a000216bc57ea8aad5b135fa0d3/zkg#L187-L191) while now we silently return nothing. This means that an operation like the following now passes ignoring the invalid path passed by the user:

```console
$ zkg --configfile /does/not/exist list
```

Maybe we want something like this?

```suggestion
        if args.configfile:
            if not file_is_not_empty(args.configfile):
                UI.error(f'invalid config file "{args.configfile}"')
                sys.exit(1)
            return str(args.configfile)
```

bbannier (commented):
Thanks for iterating. I went through all the issues and resolved most of them. I left a small number open which were about creating concrete follow-up issues.

After that I made another pass and ran a few tests, and found a few remaining regressions which I opened issues for. I believe most of these should be simple to fix.

zeekpkg/config.py
ckreibich: Thanks! I've applied a variation of this because the `normpath()` should only apply to true paths — it otherwise breaks things like `https://`.

zeekpkg/config.py
ckreibich: Yeah, thanks for noticing this. I looked at this a few ways and in the end customized `Config.update_from_file()` so its `[sources]` section (including its absence) is authoritative.

zeekpkg/config.py
ckreibich: Also a good find! I ended up checking for this right in `zkg.py`, mainly because tweaking `Config.find_configfile` quickly turns hairy due to its other callers. (And, the UI is currently used only from cli.py.)

zkg
ckreibich: Done, with very minor tweaks. I think this was a leftover from my goal of doing all those file write operations at manager instantiation time only when actually needed for the given command, but I didn't actually get to that.

zkg
```diff
@@ -3,3179 +3,46 @@
...
-                    ipkg.package.git_url,
-                    ipkg.status.current_version,
-                    False,
-                    False,
-                ),
+    if manager is None:
+        zeekpkg.UI.error("Permission problems prevent zkg from creating internal state")
+        if not args.user:
+            zeekpkg.UI.error(
+                "Consider the --user flag to manage zkg state via {home_config_dir()}/config",
```
bbannier, line 34: This was an f-string before.


```suggestion
                f"Consider the --user flag to manage zkg state via {home_config_dir()}/config",
```

bbannier (approved):
Thanks, looks good for me but for a tiny regression, opened a new issue.

I went through and again resolved all addressed threads. We still have open issues where you wanted to create issues for existing bugs/shortcomings, could you do that before moving on?

PR #235, opened 2026-07-17
Miscellaneous  cleanups

bbannier:
This PR fixes a number of unrelated minor issues, see the individual commits for more details.

### DIFF
--- a/testing/tests/info-submodules
+++ b/testing/tests/info-submodules
@@ -9,10 +9,11 @@ cp zkg.meta clonetest/zkg.meta
 
 pushd clonetest
 
-git add zkg.meta && gitcommit -v -m "Update zkg.meta"
+git add zkg.meta && git commit -v -m "Update zkg.meta"
 
 git submodule add "../$(basename $(realpath .))"
-sed 's~url.*$~url = http://github.com/zeek/A-REPO-WHICH-SHOULD-NOT-EXIST~' .gitmodules > modules && mv modules .gitmodules
+# Use a full URI so we go through the same Git machinery as a remote repo.
+sed 's~url.*$~url = file:///dev/null/A-REPO-WHICH-SHOULD-NOT-EXIST~' .gitmodules > modules && mv modules .gitmodules
 git add .gitmodules && git commit -v -m 'Add submodule'
 
 # @TEST-START-FILE zkg.meta
--- a/zeekpkg/manager.py
+++ b/zeekpkg/manager.py
@@ -1140,11 +1140,8 @@ class Manager:
 
             agg_adds = []
             agg_mods = []
-            agg_dels = []
 
             for index_file in source.package_index_files():
-                urls = []
-
                 with open(index_file) as f:
                     urls = [line.rstrip("\n") for line in f]
 
--- a/zeekpkg/manager.py
+++ b/zeekpkg/manager.py
@@ -1894,22 +1894,24 @@ class Manager:
 
         ipkg = self.find_installed_package(pkg_path)
 
-        status: PackageStatus | None = None
         if prefer_installed and ipkg:
-            status = ipkg.status
             pkg_name = ipkg.package.name
             clonepath = os.path.join(self.package_clonedir, pkg_name)
             clone = git.Repo(clonepath)
-            return _info_from_clone(clone, ipkg.package, status, status.current_version)
+            return _info_from_clone(
+                clone,
+                ipkg.package,
+                ipkg.status,
+                ipkg.status.current_version,
+            )
 
-        status = None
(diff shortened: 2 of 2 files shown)


zeekpkg/manager.py
```diff
@@ -1894,22 +1894,24 @@ def info(
...
 
-        status = None
         matches = self.match_source_packages(pkg_path)
 
         if not matches:
             package = Package(git_url=pkg_path)
 
             try:
-                return self._info(package, status, version, update_submodules)
+                return self._info(package, None, version, update_submodules)
```
awelzel, line 1914: Total nit. Would always prefer to use keyword args when passing constants, because a magic lone None would always trigger "what's the meaning of that None" when skimming code. But I'm not invested enough here :-)

testing/tests/info-submodules
```diff
@@ -12,7 +12,7 @@ pushd clonetest
 git add zkg.meta && gitcommit -v -m "Update zkg.meta"
```
awelzel: Might be worth fixing.

```
$ cat .tmp/tests.info-submodules/.stderr 
...
/home/awelzel/corelight-oss/package-manager/testing/.tmp/tests.info-submodules/info-submodules: line 12: gitcommit: command not found
```

testing/tests/info-submodules
```diff
@@ -12,7 +12,7 @@ pushd clonetest
...
 
 git submodule add "../$(basename $(realpath .))"
-sed 's~url.*$~url = http://github.com/zeek/A-REPO-WHICH-SHOULD-NOT-EXIST~' .gitmodules > modules && mv modules .gitmodules
+sed 's~url.*$~url = http://example.com/A-REPO-WHICH-SHOULD-NOT-EXIST~' .gitmodules > modules && mv modules .gitmodules
```
awelzel: > With this patch we now use an definitely non-existing repo which can be
validated without reaching out to expensive external APIs.

This does not avoid reaching out to the internet. example.com is a valid domain. It resolves to IPs and so this would now just attempt to clone something from example.com rather than github.com.

```
$ dig @8.8.4.4 example.com +short
104.20.23.154
172.66.147.243
```

After staring:

This is some sort of negative testing that would rely on cloning a submodule triggering an error?

Maybe `http://127.0.0.1:-1/A-REPO-WHICH-SHOULD-NOT-EXIST` or just `/dev/null`?

With the -1 port, git bails on the URL:
```
$ git clone clonetest x && cd x && git submodule update --init --recursive
Cloning into 'x'...
done.
Submodule 'clonetest' (http://127.0.0.1:-1/A-REPO-WHICH-SHOULD-NOT-EXIST) registered for path 'clonetest'
Cloning into '/home/awelzel/corelight-oss/package-manager/testing/.tmp/tests.info-submodules/x/clonetest'...
fatal: unable to access 'http://127.0.0.1:-1/A-REPO-WHICH-SHOULD-NOT-EXIST/': URL rejected: Port number was not a decimal number between 0 and 65535
fatal: clone of 'http://127.0.0.1:-1/A-REPO-WHICH-SHOULD-NOT-EXIST' into submodule path '/home/awelzel/corelight-oss/package-manager/testing/.tmp/tests.info-submodules/x/clonetest' failed
```

Using `/dev/null` would more look like so:
```
$ git clone clonetest x && cd x && git submodule update --init --recursive
Cloning into 'x'...
done.
Submodule 'clonetest' (/dev/null) registered for path 'clonetest'
fatal: repository '/dev/null' does not exist
fatal: clone of '/dev/null' into submodule path '/home/awelzel/corelight-oss/package-manager/testing/.tmp/tests.info-submodules/x/clonetest' failed
```

awelzel (commented):
The last commit needs some rework IMO.

testing/tests/info-submodules
bbannier: I fixed that and it changed nothing about the test, maybe worth removing in a cleanup.

testing/tests/info-submodules
bbannier: I went with a full URI so we don't accidentally have Git behave differently for clones from local paths instead of truly remote repos.

testing/tests/info-submodules
awelzel: Yeah, it's just that the changes were folded into the next commit.

awelzel (approved):

zeekpkg/manager.py
bbannier: Totally agree, but we are in the land of not just broken, but absent windows here. I'd suggest to punt on this for now to keep the changes small, and leave this for possible later cleanup. Working with a language server makes this half-bearable with signature helpers, but you are right, one should not need an LSP to understand what is going on here.

Issue #231, opened 2026-07-01
Move away from using Git to describe sources where possible


bbannier:
Currently we rely on Git to provide an interface to source files (packages or templates). We should move to a narrower interface which would allow us to support different package sources down the line, and be more conscious about where we actually require certain Git functionality. One way to implement that would be to have a pretty generic class/`Protocol` which provides generic functionality and which could be fulfilled by `git.Repo`, and use that generic approach wherever possible.

I am linking issues which could be solved more easily if we had that clean abstraction.

Issue #211, opened 2025-08-26
Consider support for Monorepos


awelzel:
Zeek distributors with very many packages are currently required to maintain packages as individual Git repositories if they want to use `zkg`.

I've observed PRs being created over 30+ repos in order to do regular package maintenance, common updates to `btest.cfg` files, updating copying/license files, or doing other common minor modifications. That feels inefficient.

While multi-repos are certainly useful if packages are authored and maintained by different developers or organizations, larger organizations may prefer to place all packages into a single repository in order to streamline and simplify updates across packages. Particular with Zeek's continuing deprecations and new language features, this is likely not uncommon.

Hypothetical, instead of having 30+ PRs to add or change a environment variable to `btest.cfg`, opening a single PR, updating 30+ `btest.cfg` files in a single repo is easier to review, approve (and also revert if something goes wrong).

There's no concrete user request here that this ticket is based on, this is mostly personal observation and impression that `zkg` precludes a monrepo-based workflow.

There's no urgency to change this. If anything, this issue can be a place for users to add their experiences and thoughts on the topic. Maybe there's also a clever workaround with git magic that could be shared here.

bbannier:
This seems like it could be useful, and is supported by other package managers, e.g.,

- Python:
  - [Packages for distribution with Poetry](https://python-poetry.org/docs/pyproject#packages)
  - [Multiple wheels in a single project with Hatch](https://hatch.pypa.io/1.12/build/)
- Rust has native support for multiple crates in a project via [workspace members](https://doc.rust-lang.org/cargo/reference/workspaces.html?highlight=workspace#the-members-and-exclude-fields)

The approach seems to be to mention contained subpackages in the meta (either explicitly or with wildcards), something which should be possible in zkg's meta as well.

The part I have not wrapped my head around is versioning. Currently the only versioning information respected by zkg are Git tags, and having multiple packages in a single repo would force a single, shared version for all packages as well. Unless we find a way to address this, e.g., by making the version explicit in a package's `zkg.meta`, zkg "monorepos" wouldn't be much different from just having a bunch of unrelated functionality in a single repo which is already possible today.

Issue #176, opened 2024-01-23
Explicitly support offline mode


ckreibich:
On machines with locked-down network connectivity `zkg`'s tendency to clone or fetch repos isn't helpful, and often not required. It'd be nice to tell it to work without remote connectivity. It'd also be good to review whether those interactions are actually needed for the commands invoked.

@ynadji, fyi!

PR #236, opened 2026-07-17
Implement support for packages not backed by Git repos

### LINKED ISSUE #158: add flag to ignore dirty local git clone

bbannier:
> [!CAUTION]
> The most offensive bit upfront: This PR was fully generated with Claude Sonnet 4.6 (all code, tests, even commit messages). This PR description is written by me. I started this out with a [design doc](https://github.com/user-attachments/files/30121205/non-git-package-source-design.md) written by me which captures some points from #231 (full of grammar errors and typos, human!). I added lots of instructions to make sure the LLM-generated commits are up to our standards and are fit for human review, and to keep it from falling back to its default, unacceptable coding style. I also requested good unit test coverage for new code, something which we currently do not have. If not automatically generated, I requested BTest integration tests as well.
> 
>Part of my instructions was to create self-contained commits which are easy to review. I did a review of all changes in this branch, so I can be and am fully accountable for anything happening here.

What this accomplishes is that we now support running operations like `zkg install` or `zkg test` against plain directories. This lifts the requirement to always commit stuff to Git before e.g., being able to run tests, and on the side also closes #158. For directories we do not clone into the local staging area, but instead work directly from it. This means that if e.g., a package requires a build step this can be shared across multiple invocations.

Since we do not require network access for such "directory packages" local directories also improve on #176 (I did not make this explicit with a flag though). The design here should also work for in the future adding support to install from tarballs which would allow users to e.g., bundle up pre-compiled artifacts without having to commit them to Git; I could imagine tarball support to extract to a tempdir, and install from that.

Since directory packages have no Git tag to determine its version, we require an explicit `version` entry in `zkg.meta` for them. Down the line this will help implement #211 since we can decouple per-package versions from Git tags on the monorepo.


timwoj (approved):
I honestly made it all the way through with zero comments. I read through the design document and the changes here appear to cover everything you wanted.

zeekpkg/manager.py
```diff
@@ -2769,16 +2770,13 @@ def test(
                 )
 
             try:
-                git_checkout(clone, version)
-            except git.GitCommandError as error:
-                LOG.warning("failed to checkout git repo version: %s", error)
+                resolution = _resolve_git_version(clone, version)
+            except Exception as error:
```
ckreibich: I pondered briefly if we could do better than `Exception` since it lists the possible exceptions below. Minor.

testing/conftest.py
```diff
@@ -1,11 +1,20 @@
...
 # Point coverage in spawned processes to global coverage configuration and state file.
 _project_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 os.environ.setdefault(
     "COVERAGE_PROCESS_START",
     os.path.join(_project_root, ".coveragerc"),
 )
 os.environ.setdefault("COVERAGE_FILE", os.path.join(_project_root, ".coverage"))
+
+
+def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
```
ckreibich, line 16: My pytest fu is weak; had to look up what this does and learned that it's a special function. A comment would help.

zeekpkg/manager.py
```diff
@@ -57,10 +57,12 @@
     PLUGIN_MAGIC_FILE_DISABLED,
     TRACKING_METHOD_BRANCH,
     TRACKING_METHOD_COMMIT,
+    TRACKING_METHOD_DIRECTORY,
     TRACKING_METHOD_VERSION,
     InstalledPackage,
     Package,
     PackageInfo,
+    PackageSnapshot,
```
ckreibich, line 66: Seeing this whole list here of 
```
    InstalledPackage,
    Package,
    PackageInfo,
    PackageSnapshot,
    PackageStatus,
    PackageVersion,
```
has me thinking that we should really look at whether these are all needed or now just artifacts of how the code happens to be laid out.

zeekpkg/manager.py
```diff
@@ -3394,6 +3437,64 @@ def _resolve_git_version(clone: git.Repo, version: str | None) -> GitResolution:
...
+    The ``version`` field in ``zkg.meta`` is mandatory since there is no
+    version control history to derive it from.
+
+    Raises:
+        ValueError: if the metadata file is missing, invalid, or has no
+            ``version`` field.
+    """
+    metadata_file = _pick_metadata_file(path)
+    metadata_parser = configparser.ConfigParser(interpolation=None)
+    invalid_reason = _parse_package_metadata(metadata_parser, metadata_file)
```
ckreibich, line 3542: I wonder if we should add an abstraction for zkg.meta. This strikes me as pretty low-level.

zeekpkg/package.py
```diff
@@ -20,6 +21,7 @@
 TRACKING_METHOD_BRANCH = "branch"
 TRACKING_METHOD_COMMIT = "commit"
 TRACKING_METHOD_BUILTIN = "builtin"
+TRACKING_METHOD_DIRECTORY = "directory"
```
ckreibich, line 24: I think this is bigger than a "tracking method", which has so far been git-specific. It's a different way of defining the whole package. I'm reading commit-by-commit so let's see if this changes, but I wonder whether the concepts should be more separate. (When working on a package, you wouldn't automatically want anything in the directory to be part of the package, for example.)

Also, it's sort of sneaking the whole concept in here, rather than focusing the commit on the snapshots.

zeekpkg/manager.py
```diff
@@ -3437,6 +3417,37 @@ def _resolve_git_version(clone: git.Repo, version: str | None) -> GitResolution:
     return GitResolution(version, tracking_method, current_hash, is_outdated)
 
 
+def _prepare_snapshot(
```
ckreibich, line 3466: I don't think this should live in manager.py since it takes a `Package` and returns a `PackageSnapshot`. So my mind goes to a method on the `Package` class.

zeekpkg/manager.py
```diff
@@ -485,6 +485,10 @@ def _read_manifest(self) -> tuple[str, str, str]:
 
             return data["script_dir"], data["plugin_dir"], data.get("bin_dir", None)
 
+    def _open_package_clone(self, package: Package) -> git.Repo:
+        """Open the on-disk git clone for an installed package."""
+        return git.Repo(os.path.join(self.package_clonedir, package.name))
+
```
ckreibich, line 498: Stuff like this strikes me as too functional/low-level. This particular function just gets a package, but somehow assumes there's a git repo in it, etc. If we had dedicated classes for git-based vs directory-based packages, you could naturally only call this for git-based ones.

zeekpkg/manager.py
```diff
@@ -2594,10 +2594,13 @@ def match_package_url_and_version(
...
             config.set("bundle", git_url, version)
 
+            if _is_directory_package(git_url):
+                return f"cannot bundle directory package {git_url}: bundling requires a Git repository"
+
             if prefer_existing_clones:
                 ipkg = match_package_url_and_version(git_url, version)
 
-                if ipkg:
+                if ipkg and _is_git_package(ipkg.status):
```
ckreibich: Here too I don't think we want separate functions. `ipkg` is an `InstalledPackage`, which already has `is_builtin()`, so it seems natural to extend it to support different types of underlying packages — perhaps just a `is_git_repo()` for now.

zkg
```diff
@@ -328,14 +328,30 @@ def is_local_git_repo_dirty(git_url: str) -> bool:
...
+    """Check whether *path* is a usable local package source.
+
+    Local paths starting with '.' or '/' must either be a clean Git repo or a
+    plain directory; everything else (URLs, SCP addresses, package source
+    names) is accepted without checks.
+    """
+    # TODO: refactor this function and its callers into a factory that maps a
+    # path to a backend kind (Git repo, plain directory, remote URL), so
+    # routing and validation are co-located and backends can be extended
+    # without touching each call site.
```
ckreibich: This! This seems to be much closer to my mental model of how this should work.

ckreibich (commented):
I've made a first real pass over this. To be clear Benjamin, I am treating this as agent code, not yours. The commits are definitely nice and manageable — very cool.

Now ... I agree that this PR gets the job done! It's impressive. But it also feels like it's it's just bolting in the functionality via `if`s and additional functions, and I can't say I like the resulting code better in the end. Low-level git code remains all over the manager ... perhaps it has to, but my hunch is we can do better. The one comment that says a more structured design is TODO is really the essence of this.

I think we could use this as a release candidate, or more likely a feature preview, that we can take to people to see how they like it. I don't think this should go into Zeek 9 before we've spent more time thinking through the design. I'd next take a fresh look at our class layout, figure out how to represent packages with different storage implementations, and then work that in.

zeekpkg/package.py
bbannier: Completely agree that this does not really belong here. I clean this up in https://github.com/zeek/package-manager/pull/237 so `DIRECTORY` and `BUILTIN` are not tracked in this enum; with that `TrackingMethod` becomes a clean enum which only models `version`, `branch`, `commit` which are all specific to the Git backend. What you see here is basically an intermediary step.

zeekpkg/manager.py
bbannier: Again agree which is, so this goes away in #237. With `TrackingMethod` only tracking Git-backed packages, a non-Git package simply has `tracking_method=None`, so this becomes trivial. The last commit in that PR then inlines the helper and deletes the function.

zeekpkg/manager.py
bbannier: Again agree. With #237 we will have a proper enum which only models Git-backed packages, so non-Git packages can never get here (prevented by the type system and mypy checking). That doesn't go as far as the full OOP tree you mentioned, but gives some structural safety. I'd argue that a full backend abstraction is overkill for two possible implementations, but happy to discuss this.

zeekpkg/manager.py
bbannier: Yeah makes sense, I narrowed this by enumerating the two possible exceptions. I think I used `Exception` here since the code called in the try block is so complicated, but I am reasonably sure we only raise the named exceptions.

testing/conftest.py
bbannier: I expanded the comment with a docs link.

zeekpkg/manager.py
bbannier: Only `PackageSnapshot` and `PackageVersion` are new here, the others were preexisting. We could try to clean this up as a follow-up, but I'd rather do that separately to not introduce more risk that needed to existing functionality.

zeekpkg/manager.py
bbannier: Like we had discussed offline a few weeks ago, this is here since it uses the existing free-standing private functions in `manager.py`. Like I mentioned then, we could try to make `Package` more encapsulated, but that would change what it does quite a bit.

zeekpkg/manager.py
bbannier: Agreed, but this is completely preexisting and orthogonal to what is done here. If you want we can create an issue and work on that next.

zeekpkg/package.py
```diff
@@ -157,6 +159,33 @@ def dependencies(
     return rval
 
 
+@dataclass
+class PackageSnapshot:
+    """A fixed, immutable view of a package at the point it enters processing.
+
+    All source-specific resolution must be completed before constructing this
```
ckreibich: A detail: I'm not sure what "source" and "resolution" mean here. I think it means the package needs to exist on disk and its current state must match that spelled out in the arguments to the constructor, right? (But that conflicts a bit with `None` for `tracking_method` being "not yet determined".

zeekpkg/manager.py
```diff
@@ -1939,14 +1941,25 @@ def _info_lookup(
...
+                )
             return _info_from_clone(
-                clone,
+                snapshot,
                 ipkg.package,
                 ipkg.status,
-                ipkg.status.current_version,
+                versions,
+                default_branch,
+                resolution.tracking_method,
```
ckreibich: This tracking method is equal to `snapshot.tracking_method`, right? I wonder whether _info_from_clone needs the argument then.

zeekpkg/manager.py
```diff
@@ -2081,11 +2088,8 @@ def package_versions(self, installed_package: InstalledPackage) -> list[str]:
         Returns:
             list of str: the version number tags.
         """
-        name = installed_package.package.name
-        assert name
-        clonepath = os.path.join(self.package_clonedir, name)
-        clone = git.Repo(clonepath)
-        return git_version_tags(clone)
+        assert installed_package.package.name
+        return git_version_tags(self._open_package_clone(installed_package.package))
```
ckreibich, line 2093: This is one of the places where I wondered why it is safe to just assume git-based packages now. Just considering the above change locally, it seems totally fine. But we can get here from the guts of `cmd_upgrade()` and `cmd_refresh()`.

Can an `InstalledPackage` only ever be git-based?

zkg
```diff
@@ -450,15 +450,21 @@ class InstallWorker(threading.Thread):
         manager: zeekpkg.Manager,
         package_name: str,
         package_version: str,
+        skip_version_validation: bool = False,
```
ckreibich, line 467: When I got to this commit again (on my second pass), it still wasn't clear to me under which circumstances you'd want this. Wouldn't it be better to just call out that there's a discrepancy, and it's not clear what to do?

zkg
```diff
@@ -328,14 +328,30 @@ def is_local_git_repo_dirty(git_url: str) -> bool:
...
 
-def check_local_git_repo(git_url: str) -> bool:
-    if is_local_git_repo_url(git_url):
-        if not is_local_git_repo(git_url):
-            print_error(f"error: path {git_url} is not a git repository")
-            return False
-        if is_local_git_repo_dirty(git_url):
-            print_error(f"error: local git clone at {git_url} is dirty")
-            return False
+def check_local_git_repo(path: str) -> bool:
```
ckreibich: I struggled with this function ... perhaps it's just the naming and the comment.

- If the path is _not_ a local git repo URL, it returns True. So that means it's a "real" URL like "https://..." or "git:...", right? That's not local, though.
- If it's a directory and there's no git repo in it, it returns True. This is the part I understand.
- But the third check is again just `not is_local_git_repo()`. What would this catch?

So perhaps this is `check_local_package_sourcetree()` or some such?

ckreibich (commented):
Okay Benjamin, I've now spent a lot more time on this and the good news is that it seems to be working as intended in my local testing. I like that it seems pretty narrow — I hadn't realized you need a local package tree that is also not a Git repo, plus presence of a version key, for the new logic to trigger. I also like the explicit disabling of such packages for bundles, which I think is pretty important at this stage.

One thing — I wonder if this is actually _too_ narrow. If the directory approach never kicks in if a local tree is a git repo, that still doesn't help much during the typical development workflow, where you _do_ work in git — you just don't want your package install to be gated on git complaints.

So I'm wondering whether we should make the tracking explicit. We could have `zkg install --directory`, perhaps, or support it in zkg.meta. If we add something like
```
[package]
tracking = git
```
as the default, and
```
[package]
tracking = directory
```
as the override, then you could work with git and have zkg ignore it. You can toggle as you see fit during development, which is cool too. I'm fairly sure this is nicer than relying on presence/absence of a version tag, which is pretty implicit. The version could also simply default to something like `devel` when not given while tracking is directory-based.

A final, minor thing: I'm not seeing much (anything?) regarding directory tracking in the output of `zkg -vvv`. It would help to make that pretty clear so there's no uncertainty about how zkg handles the package.

A few other comments in line, but I'm mostly coming around on this one. :+1:

bbannier:
Thanks for the review. Somehow I lost the actual non-Git functionality while doing a rebase, and now put it back. I still need to work through your other feedback so this is not getting ready today.

zkg
bbannier: Yeah, I cleaned this up a bit. I updated the docstring, and removed the dead code, let me know whether that helped.

zeekpkg/package.py
bbannier: I updated the class documentation, PTAL.

zeekpkg/manager.py
bbannier: > Can an `InstalledPackage` only ever be git-based?

Yes, and this is validated when constructing them. I still added an `assert` to that effect so this code is more self-contained.

zkg
bbannier: Not absolutely sure either, it felt right to allow users to opt out given that is a new consistency check. A scenario where this could become interesting would be when a single commit has multiple, conflicting Git tags; in that case a `version` field will be inconsistent for one of them. That's still pretty niche, do you think it would make sense to rip out this functionality completely?

bbannier: I removed the CLI flag, so a mismatch is now always an error; at the same time, I made the `version` field in `zkg.meta` optional.

zkg
bbannier: Like written, I already renamed the function, rewrote its docstring and dropped the third check. I am resolving this, please let me know if I should do anything else here.

bbannier:
> One thing — I wonder if this is actually _too_ narrow. If the directory approach never kicks in if a local tree is a git repo, that still doesn't help much during the typical development workflow, where you _do_ work in git — you just don't want your package install to be gated on git complaints.

I fixed one remaining issue, but I think this should work now since when working with a package we check whether the passed package name is a valid local path and if it does use the directory backend. The intention is to seamlessly enable the directory backend so that uses like `zkg test .` use the efficient approach while minimizing interference with existing workflows like `zkg install hello-world` (expected to download the `hello-world` package, not work on an local, accidentally found `hello-world/` folder). One can force the Git backend for local Git folders by using full URIs like `file:///path/too/foo` so we provide a fallback, but I do not expect that users would need this often.

> So I'm wondering whether we should make the tracking explicit. We could have `zkg install --directory`, perhaps, or support it in zkg.meta. If we add something like
> 
> ```
> [package]
> tracking = git
> ```
> 
> as the default, and
> 
> ```
> [package]
> tracking = directory
> ```
> 
> as the override, then you could work with git and have zkg ignore it. You can toggle as you see fit during development, which is cool too.

Even though this was broken in some intermediary version, my intention was always to make this implicit. My motivation was that users working with local checkouts are typically package developers working on their own packages who want `zkg test .` to "just work" (i.e., no require jumping through `git commit` hoops). We still need Git for the registry, but I feel beyond that developer users should not need to care. Making this explicit with a CLI flag seems to me like asking users to do something special for their regular use case, and we already have an opt-out with full URIs (slightly obscure, but I feel this is an obscure use as well). Capturing in `zkg.meta` seems even worse to me as it would not add much benefit, but could introduce a lot of friction, e.g., the need to change/drop `tracking` before committing and additional error scenarios.

All this said, I'd prefer to keep this implicit. WDYT?

> A final, minor thing: I'm not seeing much (anything?) regarding directory tracking in the output of `zkg -vvv`. It would help to make that pretty clear so there's no uncertainty about how zkg handles the package.

I added some log output when the directory backend is being used.

PR #237, opened 2026-07-17
Replace tracking method string constants with a `TrackingMethod` enum

bbannier:
The way we were storing the way a package came about with tracking method global strings was error prone. Any code using them was expected to figure out against which methods it should check. Since these were just unrelated global constants there was also no help from the type checker to make sure that handling was correct (forgot to check a variant?).

This PR cleans that up by introducing a proper Python enum, and then separating out Git-related tracking information from other concerns.

> [!WARNING]

### DIFF
--- a/testing/baselines/tests.builtin-info/out
+++ b/testing/baselines/tests.builtin-info/out
@@ -8,7 +8,7 @@
 		is_loaded = True
 		is_outdated = False
 		is_pinned = True
-		tracking_method = builtin
+		tracking_method = None
 	metadata (from version ""):
 		<empty metadata file>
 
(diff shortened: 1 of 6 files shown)


bbannier:
Once the enum is in place this patch presents two possible ideas:

1. `Narrow tracking method types to enforce Git-only dispatch statically`: add some annotations so mypy can typecheck things, and after reverting this
2. `Restrict TrackingMethod to Git-backed tracking strategies` which properly models this

Of course we would only merge one approach.

I think I'd prefer 2) since it is conceptually simpler loosing anything in the remodel (I believe).

ckreibich (approved):
This looks good to me Benjamin. A step in the right direction.

PR #242, opened 2026-07-22
Split long-running tests into more focused tests

bbannier:
Many of our tests check multiple behaviors at the same time. This not only makes the tests hard to follow and debug when something breaks (e.g., a test might stop before all behaviors are checked), but it also causes some tests to run for a very long time. This PR breaks up our longest running tests into more focused tests instead. I see no test suite runtime improvement from that on my machine, but IMO it still improves the overall situation.

> [!WARNING]


testing/tests/metadata-suggests-install
```diff
@@ -0,0 +1,37 @@
+# @TEST-EXEC: bash %INPUT
```
timwoj, line 1: Is this line needed? In the original test it installed `foo` directly without the edits.

testing/tests/metadata-depends-install
```diff
@@ -0,0 +1,37 @@
+# @TEST-EXEC: bash %INPUT
```
timwoj, line 1: Is this line needed? In the original test it installed foo directly without the edits.

testing/tests/metadata-suggests-bundle
```diff
@@ -1,20 +1,4 @@
-# @TEST-EXEC: zkg install foo
-# @TEST-EXEC: zkg list installed > no_suggests.out
-# @TEST-EXEC: btest-diff no_suggests.out
-
 # @TEST-EXEC: bash %INPUT
```
timwoj, line 1: Does this need to install foo? Otherwise there's nothing to bundle, right?

testing/tests/metadata-depends-bundle
```diff
@@ -1,20 +1,4 @@
-# @TEST-EXEC: zkg install foo
-# @TEST-EXEC: zkg list installed > no_depends.out
-# @TEST-EXEC: btest-diff no_depends.out
-
 # @TEST-EXEC: bash %INPUT
```
timwoj, line 1: Does this need to install foo? Otherwise there's nothing to bundle, right?

testing/tests/install-version-commit-hash
```diff
@@ -0,0 +1,33 @@
+# @TEST-EXEC: bash %INPUT
```
timwoj, line 1: Some comments in each "section" of this test might help clarify what each part is doing.

timwoj (changes requested):

testing/tests/metadata-suggests-install
bbannier: Are you thinking of the new `testing/tests/metadata-suggests-no-suggests` which contains the block before the original `bash %INPUT` line? That test does not include it. The code tested here always was after running the script.

testing/tests/metadata-depends-install
bbannier: Same here, I think you are looking for `testing/tests/metadata-depends-no-depends` (consistent naming, but urgs).

testing/tests/metadata-suggests-bundle
bbannier: Previously this ran after `zkg purge`, so there was nothing installed even before. What this did and still does is to use the `--manifest` flag which adds some package to the created bundle. Leaving that off would bundle the existing installation, but that is not what this was supposed to test.

testing/tests/metadata-depends-bundle
bbannier: Same here.

testing/tests/install-version-commit-hash
bbannier: Yeah, that would help, but also tells me that I should have broken this up further. I did that now so we have dedicated tests and no need to add sections, sorry for the churn. This forces us to copy `%INPUT` to more tests, but I think that makes sense for standalone tests (in contrast to introducing a say `tests/scripts/setup` which would couple stuff again and also make it harder to follow what is going on in each test).

`tests.install` still installs four different packages in one tests. Not sure we need section headers there, but could certainly add them. WDYT?

testing/tests/metadata-suggests-bundle
timwoj: Ah, the part about the `purge` is what I missed. Sorry for the noise.

timwoj (approved):

PR #243, opened 2026-07-22
Remove or avoid redundant work to speed up zkg

bbannier:
> [!WARNING]

This PR contains a number of unrelated cleanups which make zkg more snappy. Since I used timing of the BTest suite as a benchmark there might be more lingering improvements which are not covered by tests, or do not show up clearly. The changes fall into two categories:

1. completely avoid unneeded, expensive work, this falls more in a cleanup category
2. use light caching to avoid repeated recomputation, either during the same process invocation or by persisting slowly changing information to disk to speed up repeated zkg invocations

On my machine I see the following runtime changes of the BTest test suite:

| commit                                         | time [s] | vs prev       |
| --------------------------------------------- |------- | --------------|
| master (baseline)                             | 165.0  | --            |
| Skip source git_checkout                      | 141.7  | -23s  (-14%)  |
| Cache Manager.info()                          | 129.3  | -12s  (-9%)   |
| Disk-persist zeek --build-info                | 132.4  | +3s   (+2%)   |
| Skip submodule fetch in scratch clone         | 97.8   | -35s  (-26%)  |
| Always shallow scratch clone                  | 98.0   | +0s   (0%)    |

Like mentioned, I'd expect the commits without improvement to simply be not captured by the test suite (the increase at _Disk-persist zeek --build-info_ seems consistent with noise).

Even though all code here was generated by an LLM, the code changes seem pretty straightforward. Some new unit tests tend to go pretty heavy on mocking, but if we agree that they are still readable I think this can be justified since they are meant to capture regressions where we accidentally introduce calls of expensive operations again.

### DIFF
--- a/testing/test_manager.py
+++ b/testing/test_manager.py
@@ -1,0 +1,63 @@
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from zeekpkg.manager import Manager
+from zeekpkg.package import Package, PackageInfo
+
+
+@pytest.fixture
+def manager(tmp_path: Path) -> Manager:
+    with patch.object(Manager, "discover_builtin_packages", return_value=[]):
+        return Manager(
+            state_dir=str(tmp_path / "state"),
+            script_dir=str(tmp_path / "scripts"),
+            plugin_dir=str(tmp_path / "plugins"),
+        )
+
+
+class TestInfoCache:
+    """Manager.info() should return cached results on repeated calls."""
+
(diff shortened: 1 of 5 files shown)


timwoj (approved):
Looks great to me. I had one question that was the resolved in a subsequent commit. Much better 👍🏼

bbannier:
Merged with 06fd8b20f5db25a6525d10606821fceca3618daa, unsure why GH didn't automatically close this.

PR #251, opened 2026-08-25
Demote missing implicit script warning

### LINKED ISSUE #199: Support packages that intentionally don't have any zkg.meta-specified scripts
@J-Gras [hit a situation](https://zeekorg.slack.com/archives/CTGDG83B8/p1729250352359389) in which a package is content to only ship Zeek scripts with a contained plugin, not (also) via zkg.meta's `script_dir`. Leaving out that directive and not providing any Zeek scripts at the toplevel leads to:

```
WARNING  installing "/mypath": no __load__.zeek in implicit script_dir, skipped installing scripts
```

We have other such packages in the collection, too. Seems fine to just remove the warning, or demote it to debug level. Note that there's a related warning that would be good to confirm/check in this context, [here](https://github.com/zeek/package-manager/blob/master/zeekpkg/manager.py#L1501).

juliosuas:
## Summary

- demote the missing implicit `script_dir` message from warning to debug
- keep explicit invalid `script_dir` metadata as an installation error
- add a regression assertion for plugin-only packages without top-level scripts

Fixes #199.

## Testing

- `uv run pytest -n 0 testing/test_btest.py -k "tests.load and not dependency"`
- `uv run pytest -n 0 testing/test_manager.py testing/test_source.py --tb=short`
- pre-commit: trailing whitespace, EOF, YAML, large files, Ruff, cmake-format, and cmake-lint passed (mypy 2.1.0 hit an internal error on Python 3.13)


juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/zeek/package-manager/pull/253

PR #253, opened 2026-08-26
Demote missing implicit script warning

### LINKED ISSUE #199: Support packages that intentionally don't have any zkg.meta-specified scripts

juliosuas:
## Summary

- demote the missing implicit `script_dir` message from warning to debug
- keep explicit invalid `script_dir` metadata as an installation error
- add a regression assertion for plugin-only packages without top-level scripts

Fixes #199.

## Testing

- `uv run pytest -n 0 testing/test_btest.py -k "tests.load and not dependency"`
- `uv run pytest -n 0 testing/test_manager.py testing/test_source.py --tb=short`
- pre-commit: trailing whitespace, EOF, YAML, large files, Ruff, cmake-format, and cmake-lint passed (mypy 2.1.0 hit an internal error on Python 3.13)

---
Restored after an accidental fork deletion. Same commits as #251.


bbannier:
Do you think you could come up with a test which captures that we still emit the (now debug) message?

I assume this was fully generated by a chatbot, please disclose this in the future according to [Zeek's AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md).

timwoj:
> I assume this was fully generated by a chatbot, please disclose this in the future according to [Zeek's AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md).

Welp, I guess that last commit from "cursor[bot]" confirmed that question.

juliosuas:
Thanks for the review.

**AI disclosure** (per [Zeek AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md)): Cursor cloud agent assisted with the initial patch and the follow-up test. I reviewed and edited both before pushing.

**Debug-message test:** `testing/tests/load` now installs `i-have-no-scripts` twice — once asserting the string is absent at default log level, and once with `zkg -vvv install --force` asserting `DEBUG.*no __load__.zeek in implicit script_dir` still appears.

juliosuas:
@timwoj Fair point on the `cursoragent` author metadata — that was a slip on the follow-up commit.

I rewrote that commit so both commits on this branch are authored as me (`juliosuas`). The AI disclosure in my earlier comment still stands: Cursor cloud agent assisted; I reviewed and edited before pushing.

